PR #4, opened 2013-03-28
Additional options for command 'publish'

habibutsu:
Ability to set additional options for publish message such as 'delivery_mode' and 'headers'


michaelklishin:
This no longer can merge cleanly. Please re-submit if still relevant and we'll discuss.

PR #6, opened 2013-05-24
Guess and parse opts, node and config smarter

simmel:
Use options and overwrite them in this order:
* Defaults
* From config file
* Command line

For background, see this mailthread:
<http://thread.gmane.org/gmane.comp.networking.rabbitmq.general/19104/focus=19365>

(and sorry for taking so long, been busy on other things)


michaelklishin:
This can no longer merge cleanly. If still relevant, please re-submit and we'll discuss.

PR #32, opened 2015-04-06
Ensure that both old and new sample values are numeric

### LINKED ISSUE #30: Management DB crashes with 'badarith' if fd_used or disk_free temporarily unavailable
From an escalation:
```
 =ERROR REPORT==== 17-Feb-2015::21:14:31 ===
** Generic server rabbit_mgmt_db terminating
** Last message in was {'$gen_cast',
{event,
{event,node_stats,
<snip>
** Reason for termination ==
** {badarith,[{rabbit_mgmt_db,append_sample,6,[]},
{rabbit_mgmt_db,'-append_samples/6-lc$^1/1-0-',6,[]},
{rabbit_mgmt_db,append_samples,6,[]},
{rabbit_mgmt_db,handle_stats,6,[]} 
<snip>
```

So we get a `badarith` recording samples from `node_stats` - how can that be? `append_sample/6` guards against the new value not being a number, but does not check the old value from the `old_stats` table. Looking at `rabbit_mgmt_external_stats`, both `fd_used` and `disk_free` can return atoms when they failed to parse some external command.

Probably `append_sample/6` should just ignore the case where the old stat is not a number.

michaelklishin:
Fixes #30.

@dumbbell: ready for QA.

### DIFF
--- a/src/rabbit_mgmt_db.erl
+++ b/src/rabbit_mgmt_db.erl
@@ -745,8 +745,8 @@ append_samples(Stats, TS, OldStats, Id, Keys, Agg,
 
 append_sample(Key, Val, NewMS, OldStats, Id, Agg, State) when is_number(Val) ->
     OldVal = case pget(Key, OldStats, 0) of
-      N when is_number(N) -> N;
-      _                   -> 0
+        N when is_number(N) -> N;
+        _                   -> 0
     end,
     record_sample(Id, {Key, Val - OldVal, NewMS, State}, Agg, State),
     ok;
--- a/src/rabbit_mgmt_db.erl
+++ b/src/rabbit_mgmt_db.erl
@@ -744,8 +744,12 @@ append_samples(Stats, TS, OldStats, Id, Keys, Agg,
     end.
 
 append_sample(Key, Val, NewMS, OldStats, Id, Agg, State) when is_number(Val) ->
-    record_sample(
-      Id, {Key, Val - pget(Key, OldStats, 0), NewMS, State}, Agg, State);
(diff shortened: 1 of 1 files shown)


src/rabbit_mgmt_db.erl
```diff
@@ -744,8 +744,12 @@ append_samples(Stats, TS, OldStats, Id, Keys, Agg,
     end.
 
 append_sample(Key, Val, NewMS, OldStats, Id, Agg, State) when is_number(Val) ->
-    record_sample(
-      Id, {Key, Val - pget(Key, OldStats, 0), NewMS, State}, Agg, State);
+     case pget(Key, OldStats, 0) of
+       N when is_number(N) ->
+         record_sample(Id, {Key, Val - N, NewMS, State}, Agg, State),
+         ok;
+       _ -> ok
```
dumbbell: If the old record is not a number, shouldn't the new stat still be recorded?
```erlang
_ -> record_sample(Id, {Key, Val, NewMS, State}, Agg, State)
```

dumbbell:
Here is a simpler implementation of what I suggest:
```erlang
append_sample(Key, Val, NewMS, OldStats, Id, Agg, State) when is_number(Val) -> 
     OldVal = case pget(Key, OldStats, 0) of 
       N when is_number(N) -> N;
       _                   -> 0
     end,
     record_sample(Id, {Key, Val - OldVal, NewMS, State}, Agg, State),
     ok;
```

While here, could you please fix the indentation of the function (5 spaces instead of 4 expected)?

PR #47, opened 2015-07-22
The message 'Invalid UTF-8' is not an error.

gotthardp:
I propose the following trivial change to not scare users.

The statement appears to be an error message, but it isn't. It actually
means that the supplied binary is 'Not UTF-8' and thus needs to be encoded.
The value may be perfectly valid.


michaelklishin:
@dumbbell do you want to take over this one?

dumbbell:
Ok

dumbbell:
@gotthardp: The patch looks good to me. Could you please resubmit your patch against the `stable` branch?

PR #48, opened 2015-07-23
The message 'Invalid UTF-8' is not an error.

gotthardp:
The statement appears to be an error message, but it isn't. It actually
means that the supplied binary is 'Not UTF-8' and thus needs to be encoded.


michaelklishin:
Thank you!

PR #49, opened 2015-07-23
Split base64 encoded message property/header to multiple lines.

gotthardp:
When the message property/header contains a large binary structure, the
management web-pages become a bit ugly when displaying the message
content because the binary structure is encoded as one large string.

### DIFF
--- a/src/rabbit_mgmt_format.erl
+++ b/src/rabbit_mgmt_format.erl
@@ -100,9 +100,15 @@ utf8_safe(V) ->
         xmerl_ucs:from_utf8(V),
         V
     catch exit:{ucs, _} ->
-            Enc = base64:encode(V),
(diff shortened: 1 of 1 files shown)


michaelklishin:
@gotthardp thanks, may I ask you to take a couple of screenshots before and after the change? I feel the description is lacking context a bit.

gotthardp:
Yeah, it's not easy to describe. Have a look at the "x-amqp-1.0-properties" header.

Before: the header name is split to multiple lines and the value is so long that a horizontal scrollbar was added.
image

After: the header name is on a single line and the value is split to multiple lines. (Note that the font is not fixed-width, so each line has different length although the character count is the same.)
image

michaelklishin:
:+1:

PR #58, opened 2015-09-01
Enable resource handlers to store data into context.

gotthardp:
To process a request the Webmachine invokes several callbacks. This becomes inefficient and vulnerable to race conditions if the callbacks need to access a storage (e.g. a database). It would be nice if the handlers could store a context and load the value only once.

Example use-case: a plug-in that implements HTTP API for accessing the Last Value Cache
https://github.com/gotthardp/rabbitmq-lvc-management.git


michaelklishin:
@gotthardp note that we will likely be moving to Cowboy REST soon, so any Webmachine-specific changes need to be carefully considered. But this one seems pretty universal.

PR #59, opened 2015-09-02
Make GET /api/queues/vhost/name/get accept/provide application/json

Gsantomaggio:
We had to change the `rabbit_mgmt_test_http` since by default accepted only  [`"Authorization"`](  
https://github.com/rabbitmq/rabbitmq-management/blob/master/test/src/rabbit_mgmt_test_http.erl#L1313) as header. 

We modified [`http_upload_raw`](https://github.com/rabbitmq/rabbitmq-management/commit/bd80c1d80b4c660d42c091e5b7e9ede24d31956c#diff-0dfa2103057664e04418da6fce836d7bR1308) with the parameter `MoreHeaders` .

### DIFF
--- a/src/rabbit_mgmt_wm_queue_get.erl
+++ b/src/rabbit_mgmt_wm_queue_get.erl
@@ -17,7 +17,7 @@
 -module(rabbit_mgmt_wm_queue_get).
 
 -export([init/1, resource_exists/2, post_is_create/2, is_authorized/2,
-         allowed_methods/2, process_post/2]).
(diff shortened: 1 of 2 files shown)


michaelklishin:
Is this intended for 3.6.0?

Does it use Emacs' `erlang-mode` formatting?

Gsantomaggio:
Yes this is for 3.6.0

Yes It uses Emacs  `erlang-mode` with max 80 columns.  
But if we have to change something please let me know.

michaelklishin:
@Gsantomaggio nope, then it's all good. Thank you!

PR #71, opened 2015-10-12
Convert string to boolean in SSL config

### LINKED ISSUE #20: rabbitmqadmin misinterprets boolean settings in configfile
In

https://github.com/rabbitmq/rabbitmq-management/blob/491911cc07496bd0fade6035ae161f004ea093cd/bin/rabbitmqadmin#L344-358

the config values are read from the rabbitmqadmin.conf

In case of the ``ssl`` option the check at

https://github.com/rabbitmq/rabbitmq-management/blob/491911cc07496bd0fade6035ae161f004ea093cd/bin/rabbitmqadmin#L424

will always succeed if the option is set (e.g. ``ssl = False``)  since the string always evaluates to True.

dcorbacho:
Closes #20

### DIFF
--- a/bin/rabbitmqadmin
+++ b/bin/rabbitmqadmin
@@ -367,7 +367,10 @@ def make_configuration():
                              (options.node, options.config, error))
         else:
             for key, val in new_conf.items():
-                setattr(options, key, val)
+                if key == 'ssl':
+                    setattr(options, key, val == "True")
+                else:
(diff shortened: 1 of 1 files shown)


michaelklishin:
@dcorbacho since this can be included into `3.5.x`, can you please rebase your branch against `stable` and submit another PR for that branch?

Otherwise LGTM.

PR #73, opened 2015-10-12
Empty HTTP API responses replace by null

### LINKED ISSUE #26: Consumer utilisation for queues with no consumers is reported as "" in HTTP API responses
It should be `0.0` or `0`.

dcorbacho:
Closes #26


michaelklishin:
Thank you.

I think it can be considered a bug fix and rebased to `stable`, WDYT?

michaelklishin:
Closing in favour of #74.

Issue #50, opened 2015-08-03
Listing Many Queues is Slow


bradjorgensen:
I have a RabbitMQ cluster with about 6000 queues.  When viewing the queues through the management interface, it takes about 8-9 seconds to start sending the response and about 3 seconds to transfer the response.  The transfer time can probably be helped by supporting compression, but other than compression or making the result more concise, the transfer time is not the responsibility of the management plugin.  The time is the same whether I'm explicitly sorting the list or not.  I'm not sure how it's getting the list of queues, but I think it should be much faster.

michaelklishin:
Yes, this is a known issue.

rtraschke:
I've run some profiling on the queue display of the management UI using fprof (module rabbit_mgmt_wm_queues). A first glance points to the formatting being done by the rabbit_mgmt_format module as the main time sink (roughly ¾ of the total time taken).

I'll have a deeper look into this to see if anything reasonably obvious can be done to speed this up.

michaelklishin:
@rtraschke are you sure it's not because we load and render way more than a single page of results?

rtraschke:
@michaelklishin I noticed that all queues get transmitted back to the browser, but only a subset are rendered, there's already paging in place in the UI. But as far as I can see, the time to ship the data is dwarfed by the time to format it server side. I'll try and collate the numbers and post them later.

rtraschke:
See the Chrome timings for opening the queues pane in the presence of 12000 queues. All of the time is spent between the request getting sent and the first byte getting delivered back to the browser. That is, the time is spent mostly server side.
<img width="1222" alt="screen shot 2015-08-27 at 10 33 13" src="https://cloud.githubusercontent.com/assets/758769/9516184/4938e9aa-4ca7-11e5-8958-56a5a28f6dd8.png">

And my profiling of the server code in rabbit_mgmt_wm_queues indicates this (this was with 6000 queues, also best look at times as relative, not absolute, due to profiling overhead):
```
%                                               CNT       ACC       OWN        
[{ totals,                                     2154433,17712.489,17131.496}].  %%%

%                                               CNT       ACC       OWN        
[{ "<7278.7533.3>",                            2154433,undefined,17131.496}].   %%

% SNIP ....

{[{{rabbit_mgmt_format,strip_pids,1},          2432, 9998.131,   21.553},      
  {{rabbit_mgmt_format,queue,1},               6000, 3919.756,   51.035}],     
[...]

PR #76, opened 2015-10-22
Pagination for connections, channels, exchanges, queues

Gsantomaggio:
Fixes https://github.com/rabbitmq/rabbitmq-management/issues/50.

## Back-end

**Requests**
Server side  requests are:

    http://localhost:15672/api/queues?page=1
    http://localhost:15672/api/queues?page=1&page_size=50
    http://localhost:15672/api/queues/%2f?page=1
    http://localhost:15672/api/queues/%2f?page=1&page_size=50
    http://localhost:15672/api/queues?page=1&page_size=10&name=myqueuename
    http://localhost:15672/api/queues?page=1&page_size=10&name=^%28?=^myq%29&use_regex=true

The APIs are valid for:
* `queues`
* `exchanges`
* `connections`
* `channels`

**Params**
* `page` is the page number
* `page_size` is the  number of elements for page you want to see ( if is not defined the default value is `100`).
* `name` is case insensitive.
*  `use_regex` when is `true` enables regular expression for the param `name` 

The output is always the same:

    {
    total_count: 28758, //number elements the user can see, for example queues
    item_count: 50, //number elements present on the current page
    filtered_count:28757, //number elements filtered based on field `name` 
    page: 1,//page number
    page_size: 50,//number of element for page
    page_count: 576,// total pages 
    items: [ // list of items, for example queues
    { ...
 
***Note: the old `API(s)` still working with the old output.***

**Errors**

The pagination API handle the following errors case:

 - `page` or `page_size` are not integer values. 
    Example: `http://localhost:15672/api/queues?page=no_int_val` 
   gets:

   `"error": "bad_request", 		"reason": "…"`

    [test](https://github.com/rabbitmq/rabbitmq-management/blob/rabbitmq-management-50/test/src/rabbit_mgmt_test_http.erl#L914) 
 
 - `page` or `page_size` are out of the limit. 
    Example:`http://localhost:15672/api/queues?page=1&page_size=1000000` 
    gets:
   `"error": "…","reason": "…"`

    [test](https://github.com/rabbitmq/rabbitmq-management/blob/rabbitmq-management-50/test/src/rabbit_mgmt_test_http.erl#L916) 
 - The limit are valid, but the API trying to access to a not-exist page. 
   Example: `http://localhost:15672/api/queues?page=2&page_size=11`
  gets:
   `"error": "bad_request",
   "reason": "Page out of range…"`
     
   ( for example the user is looking a page that does not exist anymore )

   [test](https://github.com/rabbitmq/rabbitmq-management/blob/rabbitmq-management-50/test/src/rabbit_mgmt_test_http.erl#L912) 

## UI

We have changed the calls from the UI side, using the pagination.

**Auto-refresh**
We had a problem with the `autorefresh`, we solved [skipping](https://github.com/rabbitmq/rabbitmq-management/blob/rabbitmq-management-50/priv/www/js/main.js#L228) the autorefresh function when one pagination component is selected.

**Redirect**   
We had to implement the [auto-redirect](https://github.com/rabbitmq/rabbitmq-management/blob/rabbitmq-management-50/priv/www/js/main.js#L1018) page in case the page you are looking does not exist anymore. 
We wanted to implement an time control based to avoid to flood the server, in case too much redirect are called. This is a precaution.

**Dispatcher**
We modified the behavior from static to [dynamic](https://github.com/rabbitmq/rabbitmq-management/blob/rabbitmq-management-50/priv/www/js/dispatcher.js#L89). 

**Store preferences**
In order to handle the navigation, we had to store all the pagination info using [store pref](https://github.com/rabbitmq/rabbitmq-management/blob/rabbitmq-management-50/priv/www/js/main.js#L583)


priv/www/js/formatters.js
```diff
@@ -799,3 +808,32 @@ function properties_size(obj) {
     }
     return count;
 }
+
+function fmt_page_number_request(template,defaultPage){
+     if  ((defaultPage==undefined) || (defaultPage <=0))  defaultPage = 1;
```
michaelklishin: Please add spaces around `==`, `<=` and other operators.

michaelklishin:
This is far from being complete: while pagination works OK for queues, it lacks server-side filtering and there's no pagination for exchanges, channels, and connections.

I'm afraid this won't make it for 3.6.0.

PR #84, opened 2015-11-10
Delay rabbit_mgmt_db initialization

### LINKED ISSUE #81: Node fails to start after network split
The reason is that `rabbit_mgmt_db` is already running on some other node in the cluster.
As stated in https://github.com/rabbitmq/rabbitmq-management/blob/master/src/rabbit_mgmt_sup_sup.erl#L19 , death of `rabbit_mgmt_db` due to this reason should not be fatal. For this reason `rabbit_mgmt_sup` is started as a `temporary` child. But `temporary` doesn't affect (synchronous) supervisor startup process, which results in non-starting node with following message:

```
   {could_not_start,rabbitmq_management,
       {{shutdown,
            {failed_to_start_child,rabbit_mgmt_sup,
                {'EXIT',
                    {{shutdown,
                         [{{already_started,<5613.920.0>},
                           {child,undefined,rabbit_mgmt_db,
                               {rabbit_mgmt_db,start_link,[]},
                               permanent,4294967295,worker,
                               [rabbit_mgmt_db]}}]},
                     {gen_server2,call,
                         [<5200.1647.0>,{init,<5200.1645.0>},infinity]}}}}},
        {rabbit_mgmt_app,start,[normal,[]]}}}
```

I think that solution would be starting `rabbit_mgmt_sup_sup` without any children initialy. And then `rabbit_mgmt_sup` could be started using either `timer:apply_after/4` or implicitly through `rabbit_mgmt_db:safe_call/3`.

binarin:
It could prevent rabbit start in case when erlang nodes were
accidentially connected beforehand (e.g. by net_adm:ping/1).

Closes #81

### DIFF
--- a/src/rabbit_mgmt_sup_sup.erl
+++ b/src/rabbit_mgmt_sup_sup.erl
@@ -57,7 +57,8 @@ start_child() -> supervisor2:start_child( ?MODULE, sup()).
 %%----------------------------------------------------------------------------
 
 init([]) ->
-    {ok, {{one_for_one, 0, 1}, [sup()]}}.
+    timer:apply_after(0, ?MODULE, start_child, []),
+    {ok, {{one_for_one, 0, 1}, []}}.
 
 sup() ->
     {rabbit_mgmt_sup, {rabbit_mgmt_sup, start_link, []},


michaelklishin:
Thank you.

So by using a timer, we start a child asynchronously. Do you think the interval of `0` is sufficient or should we make it `> 0` (a function of `net_ticktime`, maybe?). I'm primarily curious about how you've arrived to this value.

binarin:
The meaning of 0 is "start as soon as possible", I've chosen it to stay as close as possible to original behaviour, when this child was started synchronously.

Maybe even this timer is not needed, because `rabbit_mgmt_db:safe_call/3` will also try to start this child if it is missing - but I wasn't sure that I got that part correctly.

michaelklishin:
I think your understanding of `rabbit_mgmt_db:safe_call/3` is right but relying solely on it would be pretty non-obvious. So the proposed solution sounds better to me.

michaelklishin:
@binarin after discussing this with a few folks I think your workaround is acceptable. We just need to add a couple of things (I'll take care of it):

 * More obvious checks for whether the process is already registered (there's a natural race condition there). Before we move to Raft we can't really do it entirely reliably.
 * Comments that explain the intent of this "sending a message to itself".

PR #88, opened 2015-11-26
Fix backwards sort arrow icon.

serhalp:
It's conventional in GUIs to use a triangle or arrow icon to indicate how data is currently sorted in a column – where the wide part of the triangle being at the top and the narrow part at the bottom indicates that larger rows are above (decreasing sort), and vice versa.  In other words, the direction in which the icon width increases reflects the direction in which the values in the column increase.

The current icon is contrary to most UIs out there, which can be misleading – and it just feels plain wrong:

screen shot 2015-11-26 at 3 43 57 pm

This PR swaps the two icons.


michaelklishin:
@serhalp good catch, thank you!

PR #96, opened 2015-12-23
Help popup toggle on icon (?) click.

magicpotion:
It's just a tiny fix for a thing which was annoying. It makes help popup disappear when help (?) is clicked second time. No need to move mouse around to Close popup. It also makes use of pseudo-elements to set the help (?) content.
output_koovdo


michaelklishin:
Thank you.

Issue #41, opened 2015-04-26
Statistics DB collector is a bottleneck


michaelklishin:
With a high number of stats-emitting things in a cluster (primarily connections, channels, and queues), statistics collector process gets overwhelmed and begins to use a higher-than-usual amount of RAM and (intentionally) drop stats.

We should look into parallelising it, or at least the parts that are not order-dependent.

A longer term plan is to consider using riak-core to make stats DB distributed, and aggregate results from multiple nodes. This is out of scope of this issue.

michaelklishin:
Some early thoughts on what we can do here.

We do have order-dependent bits: for example, if a connection is opened and immediately closed, we'd emit two stats events. They have to be processed in order. However, such events for separate connections can be processed in parallel. Our generic work pool won't be a good fit here: it has no hashing of any kind.

So we need work pool with hashing. We identify things by

 * Pids (connections, channels)
 * `#resource{}` records (queues, exchanges)
 * Binaries (vhosts, nodes)
 * Tuples (consumers)

Our list of worker processes will be fixed in size, so we can use either consistent hashing or modulo-based one.

ETS may still be our ultimate bottleneck. It has basic concurrency controls. We need to profile to see if all of the above would make much difference.

michaelklishin:
Another idea: we could use one collector per vhost. Aggregation then becomes more involved. Cluster-wide supervision may get more complicated as well.

gmr:
Re one collector per vhost -- that doesn't buy you much if you only use one. I'm running into rate issues at the moment with only one.

michaelklishin:
This is true. One collector per vhost could still be combined with other parallelisation strategies, though.

carlhoerberg:
Agreed, we're running hundreds of vhosts on some clusters, so for us it
would probably help a lot.

rtraschke:
Some initial results from profiling the stats db:

1. The first one is staring us in the face, but I'd not thought of it: `rabbit_mgmt_db` is a gen_server that does two things: store event data (via `handle_cast`) and query metric data (via `handle_call`). When running a profile and keeping a Mgmt UI open, the query operations completely swamp the profile data. From a cursory glance at the code, I assume this is because the query operations do quite a bit of heavy lifting to convert the event data samples into metrics.
[...]

PR #101, opened 2016-01-06
Optimise statistics DB collector

dcorbacho:
Optimisations of management DB for https://github.com/rabbitmq/rabbitmq-management/issues/41

Data is stored in ETS tables:
* one set of ETS tables per event (queue_stats, queue_exchange_stats...)
* each set contains one table per group of events (queue_msg_rates,
  deliver_get, fine_stats...) such as aggr_queue_stats_deliver_get
  (see ?AGGR_TABLES in rabbit_mgmt_metrics.hrl)
* data is then stored as a tuple (not a record) to take advantage of the
  atomic call ets:update_counter/3. The equivalent records are noted in
  rabbit_mgmt_metrics.hrl to get the position and as reference for developers
* Records are of the shape:
   {{Id, base}, Field1, Field2, ....}
   {{Id, total}, Field1, Field2, ....}
   {{Id, Timestamp}, Field1, Field2, ....}
   where Id can be a simple key or a tuple {Id0, Id1}

The process structure, vs the previous single process, is:
* One GC process per ETS table
* One querying process
* Three collectors: one collector process per priority metric (queue_stats and channel_stats) and one collector for all other metrics.


michaelklishin:
Running [rabbit-hole](https://github.com/michaelklishin/rabbit-hole) test suite on `rabbitmq-management-41` (in this repo and `management_agent`) hangs and produces an exception in the log:

```
=ERROR REPORT==== 6-Jan-2016::14:09:19 ===
webmachine error: path="/api/connections/127.0.0.1%3A55281+-%3E+127.0.0.1%3A5672"
{throw,{error,{not_a_connection_pid,<0.7530.0>}},
       [{rabbit_networking,close_connection,2,
                           [{file,"src/rabbit_networking.erl"},{line,400}]},
        {rabbit_mgmt_wm_connection,delete_resource,2,
                                   [{file,"src/rabbit_mgmt_wm_connection.erl"},
                                    {line,55}]},
        {webmachine_resource,resource_call,3,
                             [{file,"src/webmachine_resource.erl"},
                              {line,186}]},
        {webmachine_resource,do,3,
                             [{file,"src/webmachine_resource.erl"},
                              {line,142}]},
        {webmachine_decision_core,resource_call,1,
                                  [{file,"src/webmachine_decision_core.erl"},
                                   {line,48}]},
        {webmachine_decision_core,decision,1,
                                  [{file,"src/webmachine_decision_core.erl"},
                                   {line,440}]},
        {webmachine_decision_core,handle_request,2,
                                  [{file,"src/webmachine_decision_core.erl"},
                                   {line,33}]},
        {rabbit_webmachine,'-makeloop/1-fun-0-',2,
                           [{file,"src/rabbit_webmachine.erl"},{line,55}]}]}
```

I reliably can reproduce this on the topic branch but not `stable` or `master`. Other 3 test suites ((Hop)[https://github.com/rabbitmq/hop], (Ruby HTTP API client)[https://github.com/ruby-amqp/rabbitmq_http_api_client], (langohr/langohr.http)[https://github.com/michaelklishin/langohr]) pass.

dcorbacho:
That error shown in the rabbit-hole suite is caused by a call to rabbit_mgmt_db:get_connection/2, which returns the pid of a previous connection. The new create_connection event has not yet been processed by rabbit_mgmt_event_collector, so it's not reflected in the stats.

Previously, all messages would be queued in the same gen_server so create_connection would be processed before get_connection.
 
Nonetheless, the test suite still hangs if the rabbit_mgmt_db is forced to wait. I will continue investigating the issue.

dcorbacho:
It seems the connection event hasn't been processed yet, and the call finds an old PID.
The connection exists, but the management plugin doesn't find it because it depends on the stats and events are async:
https://github.com/rabbitmq/rabbitmq-management/blob/stable/src/rabbit_mgmt_wm_connection.erl#L46

Adding a 3 sec sleep call to the `ListConnections` API in `DELETE /api/connections/{name}`, makes the full test suite succeed.

The sequence is:
* The first test case in the suite `GET /overview` creates a connection
* That connection is being closed, but the event hasn't yet being processed
* The `DELETE /api/connections/{name}` test case opens a new connection
* First, it calls `ListConnections` and finds the old PID from `GET /overview`
* Next, it tries to delete that old connection
* The delete fails, because that old connection no longer exists for `rabbit_networking`. The old PID has not yet been deleted from the stats database as the event didn't arrive. However, `rabbit_networking` can see the new connection stablished and the new PID.

If commenting out that test case, there are 5 more errors. These errors are solved by adding a sleep of 1 second just after list channels https://github.com/michaelklishin/rabbit-hole/blob/master/rabbithole_test.go#L291. Thus, we have another race condition.

All these errors may be exacerbated by the new event processing, where queries and event collection are managed by different processes. Thus, queries won't wait for previous event messages. However, given the async nature of the events, this could have happened before depending on the speed of the server executing the test.

### Issues
Thus, there are three problems here: first the rabbit-hole test suite doesn't take into account that events are asynchronous. Second, `rabbit_mgmt_wm_connection` gets the PID of the connection from the stats DB, when it doens't represent the real status of rabbit. Third, I think the exception could be managed to return an error message in the API call `/api/connections/` (instead of a 500 Internal Error), as it should still be possible that the connection is closed between retrieving the PID in `rabbit_mgmt_wm_connection` and calling to `rabbit_networking`.


## Ubuntu VM 1 core - stable

A similar error can be reproduced in *stable* when running in an Ubuntu VM with a single core. When `DELETE /api/connections/{name}` lists the connections, it returns an unexpected empty list. It can be solved by adding a 1 second sleep just before.

https://github.com/michaelklishin/rabbit-hole/blob/master/rabbithole_test.go#L99

## Conclusions

To sum up, we demonstrated that there are race conditions in the test suite because of the asynchronous nature of the events. The dependency of `rabbit_mgmt_wm_connection` in the management stats is a real bug, which must be solved. The stats db cannot be used to check if a connection exists, as it is potentially out of sync with the real cluster state.

The test suite is fixed in https://github.com/michaelklishin/rabbit-hole/pull/72 It now waits for the list of connections to be empty before call `DELETE`

michaelklishin:
In a test that creates 10K and 25K queues in 2 vhosts, in this version the management UI becomes completely unresponsive. `3.5.7`, for example, is not very responsive but responds in several seconds.

Interestingly, it's not just the Queues tab, it is all of them. I don't see any particularly high CPU or scheduler load spikes, at least they are very short-lived.

michaelklishin:
For example:

```
time curl -X GET -u guest:guest http://127.0.0.1:15672/api/queues\?page\=21\&page_size\=100\&name\=\&use_regex\=false

0.00s user 0.00s system 0% cpu 17.755 total
```

michaelklishin:
Another thing that is surely not right: with only a few channels and 6K queues, on this branch
it takes more than 1 minute to list channels:

```
time curl -X GET -u guest:guest http://127.0.0.1:15672/api/channels
…
0.00s user 0.00s system 0% cpu 1:12.98 total
```

michaelklishin:
As soon as I delete all queues, the above request gets a response instantly.

michaelklishin:
I've also noticed this in a log file:

```
=ERROR REPORT==== 13-Jan-2016::00:16:19 ===
** Generic server aggr_queue_stats_queue_msg_counts_gc terminating
** Last message in was gc
** When Server state == {state,5000,#Ref<0.0.1.159401>,
                            aggr_queue_stats_queue_msg_counts,
                            {aggr_queue_stats_queue_msg_counts,
                                {{resource,<<"/">>,queue,<<"queue1774">>},
                                 total},
                                {'$1','$2'},
                                1,<<>>,[],0,0}}
** Reason for termination ==
** {badarith,[{rabbit_mgmt_stats,ets_delete_value,3,
                                 [{file,"src/rabbit_mgmt_stats.erl"},
                                  {line,383}]},
              {rabbit_mgmt_stats,gc,5,
                                 [{file,"src/rabbit_mgmt_stats.erl"},
                                  {line,293}]},
              {rabbit_mgmt_stats_gc,gc_batch,3,
                                    [{file,"src/rabbit_mgmt_stats_gc.erl"},
                                     {line,135}]},
              {rabbit_mgmt_stats_gc,handle_info,2,
                                    [{file,"src/rabbit_mgmt_stats_gc.erl"},
                                     {line,79}]},
              {gen_server2,handle_msg,2,
                           [{file,"src/gen_server2.erl"},{line,1049}]},
              {proc_lib,wake_up,3,[{file,"proc_lib.erl"},{line,250}]}]}
```

Probably entirely unrelated to the above.

michaelklishin:
@dcorbacho can you please sanity check my merge above?

dcorbacho:
@michaelklishin Done, I only found a minor issue with the merge of the tests

Issue #97, opened 2015-12-29
Validate (limit) user-provided query parameters, e.g. lengths_age


michaelklishin:
`lengths_age` and `lengths_incr ` need to be validated and capped the same way page size is.

essen:
What max values should i put?

essen:
Nevermind that. If I understand right you want to limit the number of records returned by queries that use those values, not limiting the records themselves? Ie if we request an hour of data with a sample by minutes that's 60, and perhaps we can have a hard limit of samples of 500 or something? Am I understanding right?

michaelklishin:
Currently it is possible to enter an arbitrarily high number and potentially put a serious strain on the management node. We need to limit our exposure to that.

essen:
Understood, I can reproduce (even did an OOM, oops). What I was saying in the previous comment should be at least a first step.

michaelklishin:
Fixed in #106.

PR #106, opened 2016-01-13
Limit sample ranges sent to the client

essen:
Update: I did now. (I did not change the behavior that silently dropped ranges when the input is invalid. I can change that into a proper error if required.)

Fixes https://github.com/rabbitmq/rabbitmq-management/issues/97.

### DIFF
--- a/src/rabbit_mgmt_util.erl
+++ b/src/rabbit_mgmt_util.erl
@@ -756,9 +756,8 @@ range(Prefix, Round, ReqData) ->
     Age0 = int(Prefix ++ "_age", ReqData),
     Incr0 = int(Prefix ++ "_incr", ReqData),
     if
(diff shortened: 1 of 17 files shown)


essen:
I have pushed a commit that adds a test and changes the return value in queues resources when ranges are invalid.

However I now realize there are more places that make use of ranges and more changes and tests are required to cover all resources.

essen:
OK I have added a try/catch everywhere applicable. I did not add a test everywhere applicable because it wouldn't test for much more than the test I added does (it would just test that the throw is catched). But I can add a test for every resource if you need me to.

michaelklishin:
Yes, lets do that. Even if it's the same (fairly basic) test for every resource.

essen:
On it!

essen:
Done. And good call, because I had to fix a few things.

PR #118, opened 2016-01-28
Add hash_algorithm to definitions export/import

### LINKED ISSUE #117: Include password hashing function into definitions
…and use it during definitions import.

The idea was brought up in #116 but in a different context. It's not so much about those upgrading from 3.5.x — we have a [documented procedure](http://rabbitmq.com/passwords.html) for them — but for those who will use a non-default function on 3.6.0+ and perform imports.

hairyhum:
Fixes #117


michaelklishin:
This is submitted against `master` but #117 is scheduled for `3.6.x`.

test/src/rabbit_mgmt_test_http.erl
```diff
@@ -721,6 +735,75 @@ definitions_test() ->
     rabbit_runtime_parameters_test:unregister(),
     ok.
 
+definitions_password_test() ->
+    % Import password from 3.5.x
```
michaelklishin: :+1:

PR #128, opened 2016-02-15
add base_uri option for the management API

barryib:
Sometimes, RabbitMQ are behind reverse proxy with a particular base_uri. So by using `--base-uri` option, you can use can add that particular base uri. 

The **default `base_uri` is empty**, so it'll work with standard RabbitMQ Management Plugin deployment.


michaelklishin:
This looks reasonable, thank you. Please submit the same change against `stable` if you'd like to see this in a future `3.6.x` release.

michaelklishin:
Note: we later renamed `base_uri` to `path_prefix` and ported this to `stable`.

PR #130, opened 2016-02-16
add base_uri option for the management API in 3.6.x release

barryib:
Add this option in 3.6.x release


michaelklishin:
Thanks. I will rename the option to `path_prefix` because it reflects what the flag does a bit better. I will also rename this in master.

barryib:
Ok thanks. LGTM :+1:

Issue #65, opened 2015-09-24
Export/import config on virtual host level


rtraschke:
Current export capabilities exports everything set up on the broker – users, virtual hosts, exchanges, queues, policies, etc. This does not allow for easy promotion of messaging topologies from one broker instance to another in a specific virtual host, or several virtual hosts.

It would be helpful if we could export contents of one virtual host (exchanges, queues, bindings, policies), and import it into another virtual host, i.e. the export file would contain only what’s pertinent to a virtual host rather than the entire broker instance, and would not refer to the virtual host by name (so that it can be imported into a virtual host named differently).

dcorbacho:
Done in https://github.com/rabbitmq/rabbitmq-management/pull/131

michaelklishin:
Fixed in #131, now needs porting to master.

dcorbacho:
Finished in https://github.com/rabbitmq/rabbitmq-management/pull/132

PR #131, opened 2016-02-16
Export/import config on virtual host level

dcorbacho:
Implements #65


michaelklishin:
@dcorbacho thank you. Please merge `stable` into `master`, which uses Cowboy REST.

PR #144, opened 2016-03-03
Store cluster name as binary from the HTTP API

### LINKED ISSUE #143: Cluster name can be returned as a list of integers
After running `make tests` in the plugin, `GET /api/overview` returns the following for cluster name:

```
{
    "cluster_name": [
        102,
        111,
        111
    ],
```

which breaks Hop test suite and results in a blank page in the management UI.

Assigning to @dcorbacho because she had fixed something very similar elsewhere recently :)

dcorbacho:
Fixes #143.

The problem is not in the display, but while setting the cluster name. The HTTP API was storing a string, while rabbit_nodes expects a binary (described in the spec but not enforced). With this fix the HTTP API stores a binary the same way rabbitmqctl does.

### DIFF
--- a/src/rabbit_mgmt_wm_cluster_name.erl
+++ b/src/rabbit_mgmt_wm_cluster_name.erl
@@ -46,7 +46,8 @@ to_json(ReqData, Context) ->
 accept_content(ReqData, Context) ->
     rabbit_mgmt_util:with_decode(
       [name], ReqData, Context, fun([Name], _) ->
-                                        rabbit_nodes:set_cluster_name(Name),
+                                        rabbit_nodes:set_cluster_name(
+                                          list_to_binary(Name)),
                                         {true, ReqData, Context}
                                 end).
 
--- a/src/rabbit_mgmt_wm_cluster_name.erl
+++ b/src/rabbit_mgmt_wm_cluster_name.erl
@@ -47,7 +47,7 @@ accept_content(ReqData, Context) ->
     rabbit_mgmt_util:with_decode(
       [name], ReqData, Context, fun([Name], _) ->
                                         rabbit_nodes:set_cluster_name(
-                                          list_to_binary(Name)),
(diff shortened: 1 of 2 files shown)


michaelklishin:
@dcorbacho this fails if cluster name is already a binary. Hop test that `PUT` cluster name fails for me on this branch.

michaelklishin:
```
=ERROR REPORT==== 3-Mar-2016::12:49:21 ===
webmachine error: path="/api/cluster-name"
{error,
    {error,badarg,
        [{erlang,list_to_binary,[<<"rabbit@warren">>],[]},
         {rabbit_mgmt_wm_cluster_name,'-accept_content/2-fun-0-',4,
             [{file,"src/rabbit_mgmt_wm_cluster_name.erl"},{line,50}]},
         {rabbit_mgmt_util,with_decode,5,
             [{file,"src/rabbit_mgmt_util.erl"},{line,480}]},
         {webmachine_resource,resource_call,3,
             [{file,"src/webmachine_resource.erl"},{line,186}]},
         {webmachine_resource,do,3,
             [{file,"src/webmachine_resource.erl"},{line,142}]},
         {webmachine_decision_core,resource_call,1,
             [{file,"src/webmachine_decision_core.erl"},{line,48}]},
         {webmachine_decision_core,accept_helper,1,
             [{file,"src/webmachine_decision_core.erl"},{line,616}]},
         {webmachine_decision_core,decision,1,
             [{file,"src/webmachine_decision_core.erl"},{line,521}]}]}}
```

The test [is here](https://github.com/rabbitmq/hop/blob/master/src/test/groovy/com/rabbitmq/http/client/ClientSpec.groovy#L1032). README has [a node on running tests with Gradle](https://github.com/rabbitmq/hop#running-tests).

michaelklishin:
OK, seems to be good now.

Issue #142, opened 2016-03-02
Make it easier to differentiate between Delete and Purge buttons on the queue page


dkoen:
We had an issue with permissions where someone meant to purge a queue and accidentally clicked the delete button which caused us some minor problems. Would it be possible to make the difference between the buttons more clear? As it stands it's pretty hard to tell (popup withstanding). 

From previous: https://github.com/rabbitmq/rabbitmq-server/issues/670

michaelklishin:
I don't have a suggestion for how the two can be distinguished better — both are destructive operations — and the Delete button has a confirmation.

I really fail to see how #146 would make anything clearer. The real problem here is probably that the Purge button doesn't have a confirmation, even though it's a destructive operation. We could also rename Purge to Purge Messages, at least that would reflect that the queue itself won't be removed.

michaelklishin:
@Gsantomaggio do you have an opinion on this?

Gsantomaggio:
About the #146 I think that won't help.
In order to be started we would modify also `exchange` section  adding `delete exchange`,  `admin`  section adding `delete user / vhost` etc.

The improvement I see are:

1. Add `Purge` confirmation ( as suggest by @michaelklishin )
2. Change the color in `red` for the delete operations, somethings like `Danger Zone` on github settings.

jackar:
@Gsantomaggio we've had multiple independent teams/individuals accidentally delete queues because of the placement of these buttons. See [fitt's law](https://en.wikipedia.org/wiki/Fitts%27s_law) for an HCI theory on distance between logical targets. 

It's one thing to write the front-end layer and know the placement of the buttons and another to be a new user. Delete/purge is unclear. It could mean delete/purge all messages, accounting for their synonymous definitions, which could cause a user to click delete without having gazed to the other side of the screen where the purge button appears. On the other hand, it could also be interpreted in the way that was intended.

Ultimately, this issue revolves around semantics, I'm not sure how helpful coloration will be and confirmation for purge is a solution without a problem.

bdshroyer:
Addressed in [PR #142](https://github.com/rabbitmq/rabbitmq-management/pull/147).

PR #146, opened 2016-03-03
better naming conventions wrt delete / purge

jackar:
https://github.com/rabbitmq/rabbitmq-management/issues/142

This will prevent the accidental deletions of a queue. Frankly, the position of delete/purge should also change, but I'll open another issue for that soon. The current state has delete on the far left and purge on the far right, a bit far away to notice it on first glance if you're new to rabbitmq and you've been asked to purge the queue.

### DIFF
--- a/priv/www/js/tmpl/queue.ejs
+++ b/priv/www/js/tmpl/queue.ejs
@@ -295,14 +295,14 @@
       <input type="hidden" name="vhost" value="<%= fmt_string(queue.vhost) %>"/>
       <input type="hidden" name="name" value="<%= fmt_string(queue.name) %>"/>
       <input type="hidden" name="mode" value="delete"/>
-      <input type="submit" value="Delete" />
+      <input type="submit" value="Delete Queue" />
     </form>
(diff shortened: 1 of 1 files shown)


michaelklishin:
I appreciate your interest in contributing but is it really an improvement? I don't think so. I don't have a suggestion for how the two can be distinguished better — both are destructive operations — and the Delete button has a confirmation.

I really fail to see how "Delete Queue" is easier to tell from "Purge Queue" than "Delete" is from "Purge".

michaelklishin:
See some specific ideas on what would actually improve things a bit in #142.

michaelklishin:
@jackar I really really fail to see how moving the buttons together would make things safer.

jackar:
@michaelklishin 'Delete Queue' and 'Purge Queue' clarify meaning by avoiding the misinterpretation of the closer button 'Delete' to be understood as 'Delete Messages'. See https://github.com/rabbitmq/rabbitmq-management/issues/142.

PR #147, opened 2016-03-04
improvement w.r.t. wording

jackar:
https://github.com/rabbitmq/rabbitmq-management/issues/142

### DIFF
--- a/priv/www/js/tmpl/queue.ejs
+++ b/priv/www/js/tmpl/queue.ejs
@@ -302,7 +302,7 @@
       <input type="hidden" name="vhost" value="<%= fmt_string(queue.vhost) %>"/>
       <input type="hidden" name="name" value="<%= fmt_string(queue.name) %>"/>
       <input type="hidden" name="mode" value="purge"/>
-      <input type="submit" value="Purge Queue" />
+      <input type="submit" value="Purge Messages" />
     </form>
   </div>
 </div>
--- a/priv/www/js/tmpl/queue.ejs
+++ b/priv/www/js/tmpl/queue.ejs
@@ -295,14 +295,14 @@
       <input type="hidden" name="vhost" value="<%= fmt_string(queue.vhost) %>"/>
       <input type="hidden" name="name" value="<%= fmt_string(queue.name) %>"/>
       <input type="hidden" name="mode" value="delete"/>
-      <input type="submit" value="Delete" />
+      <input type="submit" value="Delete Queue" />
     </form>
 
     <form action="#/queues" method="delete" class="inline-form-right">
       <input type="hidden" name="vhost" value="<%= fmt_string(queue.vhost) %>"/>
       <input type="hidden" name="name" value="<%= fmt_string(queue.name) %>"/>
       <input type="hidden" name="mode" value="purge"/>
-      <input type="submit" value="Purge" />
+      <input type="submit" value="Purge Queue" />
     </form>
   </div>
 </div>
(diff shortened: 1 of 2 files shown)


priv/www/js/tmpl/queue.ejs
```diff
@@ -295,14 +295,14 @@
       <input type="hidden" name="vhost" value="<%= fmt_string(queue.vhost) %>"/>
       <input type="hidden" name="name" value="<%= fmt_string(queue.name) %>"/>
       <input type="hidden" name="mode" value="delete"/>
-      <input type="submit" value="Delete" />
+      <input type="submit" value="Delete Queue" />
```
michaelklishin: I think it's obvious what the button deletes. Can we please back out this specific change?

priv/www/js/tmpl/queue.ejs
```diff
@@ -295,14 +295,14 @@
...
     </form>
 
     <form action="#/queues" method="delete" class="inline-form-right">
       <input type="hidden" name="vhost" value="<%= fmt_string(queue.vhost) %>"/>
       <input type="hidden" name="name" value="<%= fmt_string(queue.name) %>"/>
       <input type="hidden" name="mode" value="purge"/>
-      <input type="submit" value="Purge" />
+      <input type="submit" value="Purge Messages" />
```
michaelklishin: :+1:

jackar:
image

priv/www/js/tmpl/queue.ejs
jackar: I'm inclined to disagree - just because you and I know what the implication is, that doesn't mean new developers do. I don't see a problem with increasing clarity - it's the only time when the Delete expand shows two buttons. Unless someone's already had experience deleting an exchange, they might not catch on. It's more of a safety measure.

bdshroyer:
A quick unscientific poll of the RMQ users in my office didn't show any confusion about the difference between the 'delete' and 'purge' options, even among new users. Most of their accidental queue deletions were driven by haste or habit, and were generally caught by the confirmation window. There was, however, some support for moving the two buttons closer together, and possibly reordering them so that Purge (the less destructive option) is on the left.

@jackar Have you field-tested this PR with your group? What were the results?

michaelklishin:
This has sparked quite a bit of internal debate. The conclusion seems to be to merge this and then introduce a confirmation dialog for the purge button.

bdshroyer:
I've collected some user and designer feedback over the last couple of weeks, and I'll try to summarize the results below. TL;DR: We're merging the change and soliciting feedback.

The user feedback I've received suggests that confusion about deletes vs. purges was not widespread. However, changing the button labels to 'Purge Messages' and 'Delete Queue' was supported by both users and designers.

The placement of the buttons was a little bit more divisive. The users I spoke with reported that most of their accidental deletions came from hasty action (i.e., clicking Delete from force of habit, or because they're working too fast and Delete is more conveniently placed). Some of the users wanted to keep them on opposite sides of the page to make the act of choosing one over the other more deliberate. Others preferred to move them closer together, arguing that having both of them in sight was a reminder that there were two options.

I spoke with a designer and an ex-designer PM, who both liked that the Purge button was easier to find on the left side. They were both a little uneasy about its proximity to Delete, though--as one pointed out, putting two destructive operations next to each other is unlikely to reduce accidents.

All that said, the general consensus was that this is not a change likely to cause problems for anyone. Given this, we're going to merge the change and collect as much user feedback as we can. The designers also suggested some other changes that will find their way into further Github issues.

@jackar, thank you for bearing with us on this.  Can you let us know how this change works out for your team?


Thanks,

Brandon

PR #150, opened 2016-03-09
Calculate average operation times from accumulated time and count

dcorbacho:
Part of https://github.com/rabbitmq/rabbitmq-server/issues/557
Depends on https://github.com/rabbitmq/rabbitmq-management-agent/pull/13

Average operation times were miscalculated by handling them like any other system metric. These two PR solve this by reporting the raw data and calculating averages during the queries. Samples are reported as instant rates, so they truly reflect the behaviour of the system during that point in time.

### DIFF
--- a/include/rabbit_mgmt_metrics.hrl
+++ b/include/rabbit_mgmt_metrics.hrl
@@ -36,10 +36,10 @@
 
 -define(COARSE_NODE_STATS,
         [mem_used, fd_used, sockets_used, proc_used, disk_free,
-         io_read_count,  io_read_bytes,  io_read_avg_time,
-         io_write_count, io_write_bytes, io_write_avg_time,
-         io_sync_count,  io_sync_avg_time,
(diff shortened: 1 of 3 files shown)


michaelklishin:
I'm observing the following when running rabbithole tests in a row (they do not result in a lot of seeks, of course):



This looks a bit off.

michaelklishin:
I can repeatedly reproduce the issue above with `PeftTest` or from a REPL, after multiple rebuilds (at first I thought my agent build could be out of date). As soon as the "spike" reaches the left hand side, it stays there forever and the rest of the chart stays flat.

dcorbacho:
Got it, that seems to happen when the load drops. Without looking at the code, I think there is a flaw on the new samples design. The first sample taken has no previous timestamp to compare with, so it get all the history from the start-up of the node. Thus, if it had a high load you end up with that.

Issue #75, opened 2015-10-21
Wrong JSON data type for empty `publish-properties` in /api/parameters/shovel


smee:
If I create a dynamic shovel via pushing to `/api/parameters/shovel` with a `publish-properties` attribute like this
```
 curl -i -u guest:guest -H "content-type:application/json" -XPUT \
-d'{"value":{"dest-exchange":"amq.direct","src-queue":"test","dest-uri":"amqp://", "src-uri":"amqp://","publish-properties":{"user_id":"guest"}}}' \
http://localhost:15672/api/parameters/shovel/%2f/test1
```
I can see at `/api/parameters/shovel/%2f/test1` that the attribute `publish-properties` is a JSON object. If I instead create a queue with an empty `publish-properties` like

```
 curl -i -u guest:guest -H "content-type:application/json" -XPUT \
-d'{"value":{"dest-exchange":"amq.direct","src-queue":"test","dest-uri":"amqp://", "src-uri":"amqp://","publish-properties":{}}}' \
http://localhost:15672/api/parameters/shovel/%2f/test1
```
the data type of this field is a JSON array instead. Is it possible to either not show the property or use an empty object?

michaelklishin:
This was fixed in several places before,
returning an empty array is possible.

smee:
You mean only an empty array? Not an empty object?

michaelklishin:
An empty object, sorry. The issue here is that proplists are lists in Erlang and so empty proplists are serialised as empty lists unless mochijson2 is instructed otherwise.

michaelklishin:
This belongs to `shovel-management` but lets keep it here.

michaelklishin:
We've evaluated this and there is an issue with Erlang JSON libraries prior to 17.x: they use lists to express what is objects in JSON. So we can't reliably tell an empty list from an empty object. In master we will switch to maps. For `3.6.x`, we will look at what we can do.

essen:
Just sent two PRs with a proposal for a fix for this on stable. It's dirty really, and I have not run the full test suites to make sure I didn't break anything. Looking to see if that sort of fix is OK.

PR #155, opened 2016-03-11
Add a hack to properly return an empty JSON object for shovel

essen:
Part of proposal to fix https://github.com/rabbitmq/rabbitmq-management/issues/75 on stable

### DIFF
--- a/src/rabbit_mgmt_wm_parameter.erl
+++ b/src/rabbit_mgmt_wm_parameter.erl
@@ -52,7 +52,8 @@ resource_exists(ReqData, Context) ->
      end, ReqData, Context}.
 
 to_json(ReqData, Context) ->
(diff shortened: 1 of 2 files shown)


michaelklishin:
This has conflicts, please resolve.

essen:
Should be good. Sorry I didn't see the message earlier.

michaelklishin:
@essen can you please look into a PR for master soon? we have a script that spams us if stable isn't merged into master in any of the repos. Thank you.

essen:
The "fix" for master is the switch to a JSON library that uses maps. I started but am nowhere close to finish. You may also apply this patch to master directly (or I can add a PR with it) but that's basically a waste of time.

michaelklishin:
We can merge stable into master and undo all the changes for now, then transition to a new shared JSON library. WDYT?

essen:
Fine with me, just weird. :-)

Issue #152, opened 2016-03-09
rabbitmqadmin should provide a way to disable peer verification


michaelklishin:
I know, I know. This is not a good idea but sometimes it's not entirely unreasonable to do that with self-signed certificates, and in particular CAs. See also #151.

Issue #151, opened 2016-03-09
rabbitmqadmin should provide a way to specify trusted CAs


michaelklishin:
Currently we (implicitly) load the list of system CAs and Python's ssl module seems to assume that the client's own CA should not be trusted :/

See [Python ssl docs](https://docs.python.org/2/library/ssl.html) for more info.

carlhoerberg:
This seems to have broken normal TLS, eg. when the client just should verify the server's cert and is not providing a client cert. Will submit a patch that will support both cases.

PR #160, opened 2016-03-18
HTTPS support improvements around peer verifications

### LINKED ISSUE #149: rabbitmqadmin should not use deprecated HTTPSConnection methods
E.g. we do not use [ssl.SSLContext.load_cert_chain](https://docs.python.org/2/library/ssl.html#ssl.SSLContext.load_cert_chain) as [HTTPSConnection docs](https://docs.python.org/2/library/httplib.html) suggest.

michaelklishin:
Fixes #149, #151, #152. This also disables SSLv3 and requires peer certificate to be present (effectively always the case with RabbitMQ server) as drive-by changes.

CC @jerrykuch.


michaelklishin:
I will QA this in a bit. The changes are not actually by me, I just committed & pushed them.

PR #162, opened 2016-03-22
Enable read-only access to policies for management and monitoring users

### LINKED ISSUE #156: View policies for users tagged under 'management' and 'monitoring' roles
As of March 2016, policies are visible (and also create, update and delete) only for those users who are tagged under the 'policymaker' and 'administrator' roles. Please consider the 'management' and 'monitoring' roles to allow viewing policies but not updating/deleting them.

hairyhum:
Fixes #156

### DIFF
--- a/priv/www/js/tmpl/policies.ejs
+++ b/priv/www/js/tmpl/policies.ejs
@@ -27,7 +27,11 @@
 <% if (vhosts_interesting) { %>
      <td><%= fmt_string(policy.vhost) %></td>
 <% } %>
+<% if (user_policymaker) { %>
      <td><%= link_policy(policy.vhost, policy.name) %></td>
+<% } else { %>
+     <td><%= fmt_string(policy.name) %></td>
+<% } %>
      <td><%= fmt_string(policy.pattern) %></td>
      <td><%= fmt_string(policy['apply-to']) %></td>
      <td><%= fmt_table_short(policy.definition) %></td>
@@ -42,6 +46,7 @@
   </div>
   </div>
 </div>
+<% if (user_policymaker) { %>
 
 <div class="section-hidden">
   <h2>Add / update a policy</h2>
@@ -130,3 +135,4 @@
     </form>
   </div>
 </div>
+<% } %>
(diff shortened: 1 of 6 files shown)


priv/www/js/tmpl/policies.ejs
```diff
@@ -27,7 +27,11 @@
 <% if (vhosts_interesting) { %>
      <td><%= fmt_string(policy.vhost) %></td>
 <% } %>
+<% if (user_policymaker) { %>
```
michaelklishin: Please add the missing `is`. I do realise that `vhosts_interesting` and probably a few other places have the same issue but we shouldn't make it worse.

test/src/rabbit_mgmt_test_http.erl
```diff
@@ -1650,7 +1650,7 @@ policy_permissions_test() ->
 
     http_put("/users/admin",  [{password, <<"admin">>},
                                {tags, <<"administrator">>}], [?CREATED, ?NO_CONTENT]),
-    http_put("/users/mon",    [{password, <<"monitor">>},
+    http_put("/users/mon",    [{password, <<"mon">>},
```
michaelklishin: Does this value need to change?

michaelklishin:
If we expose policies, we might as well expose parameters.

test/src/rabbit_mgmt_test_http.erl
hairyhum: Yes, because in `http_get` password should be same as username.

hairyhum:
Should we show parameter based pages like federation upstreams and shovel management if we expose parameters?

michaelklishin:
Those come from separate plugins. I'd say we should but please file separate issues for federation-management and shovel-management.

uvzubovs:
Please be careful and not expose upstreams because URIs contain federation user password.

uvzubovs:
Would you please help me understand the status of this item? Was it completed or deferred? I am trying to get policy info in 3.6.5, and it seems that I still need policymaker tag. Thanks!

dcorbacho:
Users with management/monitoring tags are allowed to list policies and their details using `/api/policies`, but `/api/policies/vhost/name` is restricted to policymakers as that URI allows to modify (create/delete) the policy.

Issue #157, opened 2016-03-14
Management API to return CORS headers


epretha:
Producing a monitoring or alternative management GUI that uses HTTP API from JavaScript is not currently possible via XHR because HTTP API does not return CORS headers.

michaelklishin:
We had this request in the past, https://github.com/rabbitmq/rabbitmq-management/issues/3, which was fixed for Web STOMP but not HTTP API.

uvzubovs:
Thank you for considering this. For now, we will have to use IE.

essen:
Not be a problem, one question though: should this be configurable or always-on? And if configurable, what should be the default?

michaelklishin:
Unless there are reasons to disable it, probably always on.

Headers can be stripped by reverse proxies as a workaround.

Are there any known downsides?

essen:
None that I can think of at the moment.

essen:
OK after reading more on the subject I think we should not enable it by default. My fear is that people may have a management node exposed to the Internet, and the combination of "allowing everything" and "exposed to everyone" can lead to CSRF attacks.

In the use case here we want to allow only on a specific origin (company-specific management GUI) or origins, so users should configure the allowed origin(s) explicitly to only allow requests from that GUI. If not configured or value is undefined, then all origins are rejected (the same as today). Note that users can still configure it to allow everything (can be convenient, but we should stress out not to do this unless behind a firewall).

The other interesting configuration value is the max-age for the preflight request cache. I also need to find a good default for this. I will look at what is usually done.

In the future it might also be necessary to handle `Access-Control-Expose-Headers` but I don't think it will matter for the time being.

michaelklishin:
@essen thank you, we can start with "disabled by default", that's fine. I doubt many management UI instances are exposed on the public Internet: default credentials are limited to localhost-only access by default, for example. But I see where you're coming from.

uvzubovs:
I don't have preference one way or another, as long as we can enable CORS headers somehow. It just seems that exposing management UI to public internet is rather naïve as Rabbit does not implement credential locking on certain number of failed login attempts.

essen:
I have just pushed a branch implementing CORS support on stable. Work remain to be done for master and for the documentation.

michaelklishin:
@essen can you please look into a PR for master soon? we have a script that spams us if stable isn't merged into master in any of the repos. Thank you.

essen:
[...]

PR #167, opened 2016-04-04
Add CORS support

essen:
Two new options: cors_allow_origins and cors_max_age.

* cors_allow_origins: list of origins allowed, for example
  ["http://rabbitmq.com", "http://pivotal.com"];
  or ["*"] to allow everything.

* cors_max_age: time in seconds that clients may cache
  preflight requests. It defaults to 30 minutes.

Stable version for https://github.com/rabbitmq/rabbitmq-management/issues/157


essen:
Not sure what the Travis errors are about.

PR #168, opened 2016-04-04
Add exchange to exchange hint

sieben:
It could be useful to add this indication to help beginners to understand how exchange to exchange could be configured.


michaelklishin:
Thank you for your time.
MI'm sorry but I feel this clarifies little. It already says "two exchanges" and the table cell updated seems to be the one listing exchanges.

Issue #163, opened 2016-03-29
Report more TLS-related values via HTTP API


pickettphil:
Using /api/overview or /api/nodes, it would be useful to see all ssl_options reported.  In previous versions, we saw something like the following reported:

`{"node":"rabbit@localhost","description":"RabbitMQ Management","path":"/","port":"15672","ssl":"true","ssl_opts":{"cacertfile":"/etc/rabbitmq/ca.pem","certfile":"/etc/rabbitmq/cert.pem","keyfile":"/etc/rabbitmq/key.pem"}}`

Our configuration also specified verify, fail_if_no_peer_cert, versions ,.... As a result, we don't know what Rabbit does if any of these were overridden from command line.

I tested with both 3.5.7 and 3.6.1 and don't see any ssl_options reported at all.

essen:
@pickettphil I can't reproduce so far. I see the options as expected on current stable when starting a single node with management enabled.

Can I see the relevant part of the configuration file? Is this with only management started or is there other nodes? Is the correct configuration file picked up at startup?

michaelklishin:
@essen are trying with the tip of stable or 3.6.1?

essen:
@michaelklishin Stable after doing "make up".

michaelklishin:
Please try with 3.6.1 to compare ;)

essen:
No problem with the 3.6.1 tag (should I try the release directly?)

michaelklishin:
There is no harm in trying `3.6.1` from a release tarball but I expect there to be no difference. We should wait for more details from @pickettphil.

pickettphil:
I have the following rabbitmq.config:

[
  {rabbit, [
     {ssl_listeners, [5671]},
     {ssl_options, [{cacertfile,"/opt/pivotal/ssl/testca/cacert.pem"},
                    {certfile,"/opt/pivotal/ssl/server/cert.pem"},
                    {keyfile,"/opt/pivotal/ssl/server/key.pem"},
                    {verify,verify_peer},
                    {fail_if_no_peer_cert,false}]}
   ]}
].

and am running RabbitMQ 3.6.1 on Erlang 18.2 on CentOS 7. I see the
following:

http://localhost:15672/api/overview
===========================
{"management_version":"3.6.1","rates_mode":"basic","exchange_types":[{"name":"direct","description":"AMQP
direct exchange, as per the AMQP
specification","enabled":true},{"name":"headers","description":"AMQP
[...]

PR #169, opened 2016-04-05
Add socket options to listener information (DO NOT MERGE)

essen:
Part of https://github.com/rabbitmq/rabbitmq-management/issues/163


michaelklishin:
See https://github.com/rabbitmq/rabbitmq-common/pull/82. This involves a database schema change so it cannot go into stable.

PR #172, opened 2016-04-06
Display user-provided connection name, if any

hairyhum:
If client provided client property `connection_name`, it will be displayed in management GUI.

Part of rabbitmq/rabbitmq-server#104.


michaelklishin:
@hairyhum we should not completely replace the name we have. For several reasons:

 * TCP tuple-based ID is used to close connections or get specific information for it. We absolutely cannot break this for existing systems.
 * User-provided names are not guaranteed to be unique. TCP tuple-based names are (per node).
 * #500 still uses TCP tuple-based names heavily.

So the user-provided name, when present, should be listed in a separate column (and on the connection page), not replace the name we have.

Issue #161, opened 2016-03-22
Add a new endpoint that returns memory breakdown


michaelklishin:
Like `rabbitmqctl status`.

michaelklishin:
Note: we already provide access to this at `GET /api/nodes/{node}` if the `memory` query parameter is set to `true`. However, it is not documented and easy to find. We should introduce a new endpoint and document it: `GET /api/nodes/{node}/memory`, for example.

More ideas: relative memory use breakdown (in % of total), e.g. as a separate endpoint such as `GET /api/nodes/{node}/memory/relative`.

dcorbacho:
Done in #181

PR #181, opened 2016-04-15
Add endpoints for memory: absolute and relative

dcorbacho:
Add endpoints for memory: absolute and relative
Solves #161 

@michaelklishin This PR is missing the tests for relative memory, as I need confirmation of:
1. Is it fine to report relative memory as an integer value?
2. If memory usage is very low, but not 0, should it be reported as 0 or 1? Meaning that the component is indeed doing something.

### DIFF
--- a/src/rabbit_mgmt_dispatcher.erl
+++ b/src/rabbit_mgmt_dispatcher.erl
@@ -41,6 +41,7 @@ dispatcher() ->
      {["cluster-name"],                                            rabbit_mgmt_wm_cluster_name, []},
      {["nodes"],                                                   rabbit_mgmt_wm_nodes, []},
      {["nodes", node],                                             rabbit_mgmt_wm_node, []},
+     {["nodes", node, "memory"],                                   rabbit_mgmt_wm_node_memory, []},
      {["extensions"],                                              rabbit_mgmt_wm_extensions, []},
      {["all-configuration"],                                       rabbit_mgmt_wm_definitions, []}, %% This was the old name, let's not break things gratuitously.
      {["definitions"],                                             rabbit_mgmt_wm_definitions, []},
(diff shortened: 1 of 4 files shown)


michaelklishin:
@dcorbacho I'd say integer values (as %) is good enough. Reporting a value as 0 would indeed be weird, so I'm OK with using 1% for those cases.

michaelklishin:
The relative endpoint returns a total value that seems absolute. If other numbers are meant to add up to 100, I'd say we should return 100 for `total`:

```
{
    "memory": {
        "atom": 1,
        "binary": 1,
        "code": 24,
        "connection_channels": 0,
        "connection_other": 1,
        "connection_readers": 0,
        "connection_writers": 0,
        "mgmt_db": 1,
        "mnesia": 1,
        "msg_index": 1,
        "other_ets": 2,
        "other_proc": 16,
        "other_system": 54,
        "plugins": 3,
        "queue_procs": 1,
        "queue_slave_procs": 0,
        "total": 100594608
    }
}
```

@dcorbacho WDYT?

dcorbacho:
I think that you are right. I'll fix it now

gerhard:
We've just come across this part of rabbitmq/rabbitmq-server#1259 and noticed 2 things:

1. Relative memory usage adds up to `106%` in a just started RabbitMQ node and `115%` in one of the 3.6 LRE nodes. Would it be worth returning fractions with single precision rather than rounded integers? Single precision rounding will skew the total by at most 1.7% (17 values * 0.1)
1. The `/api/nodes/NODE/memory` API endpoint is not documented in `/api`. Should it be?

michaelklishin:
@gerhard we can switch to floats with, say, 2 decimal places. @dcorbacho @kjnilsson WDYT?

gerhard:
That would be perfect. Can this ship in 3.6 or is it a breaking change?

kjnilsson:
2 decimal places sounds resonable.

michaelklishin:
@gerhard most JSON parsers should parse integers much like floats. Computations on floats can work differently depending on the language. I think it's acceptable given other changes around VM memory consumption calculations.

PR #207, opened 2016-05-23
Don't log table name and interval as a tuple

### LINKED ISSUE #206: Table name is logged as a tuple
When a node starts, it logs a bunch of entries about ETS table GC processes:

```
=INFO REPORT==== 21-May-2016::15:03:57 ===
Statistics garbage collector started for table {aggr_connection_stats_coarse_conn_stats,
                                                5000}.
```

In the `3.6.2` cycle it used to log just the table name but then an interval was introduced without updating log entry format.

kjnilsson:
Fixes #206

### DIFF
--- a/src/rabbit_mgmt_stats_gc.erl
+++ b/src/rabbit_mgmt_stats_gc.erl
@@ -66,7 +66,7 @@ start_link(Table) ->
 
 init([Table]) ->
     {ok, Interval} = application:get_env(rabbit, collect_statistics_interval),
-    rabbit_log:info("Statistics garbage collector started for table ~p.~n", [{Table, Interval}]),
+    rabbit_log:info("Statistics garbage collector started for table ~p with interval ~p.~n", [Table, Interval]),
     {ok, set_gc_timer(#state{interval = Interval,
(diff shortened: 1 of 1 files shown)


src/rabbit_mgmt_stats_gc.erl
```diff
@@ -66,7 +66,7 @@ start_link(Table) ->
 
 init([Table]) ->
     {ok, Interval} = application:get_env(rabbit, collect_statistics_interval),
-    rabbit_log:info("Statistics garbage collector started for table ~p.~n", [{Table, Interval}]),
+    rabbit_log:info("Statistics garbage collector started for table ~p at interval ~p.~n", [Table, Interval]),
```
michaelklishin: This makes it sound like the sample GC process is started periodically while the interval defines how often GC "runs" are performed. What do you think about re-wording it a bit like "Statistics garbage collector started for table ~p with run  interval ~p."?

kjnilsson: yes agree 'with' is clearer.  will updated

Issue #216, opened 2016-06-06
Configurable toggle or retention of per-channel stats


gmr:
In an attempt to figure out what the core issue is with bottlenecks for us in the management plugin rates I created a stress testing tool. I believe the issue I am looking into is related to #214.

The tool @ https://github.com/gmr/mgmt-stress opens up N connections, then every 100 MS, it opens one or more channels and publishes a random number of messages. 

With the default settings of the app, if the management plugin rates mode is not enabled, it can easy handle the channel turnover and message velocity. 

When rates mode is set to basic, the memory quickly outgrows the docker container. If I change the channel count to 1, the app instead uses a single channel per connection that is opened on startup instead of creating a new channel every 100ms, publishing it, and closing it.

In testing, I also noticed that even after the channels go away, the memory that was used for the channels that churn does not go away.

I was wondering if a quick solution for this would be to be able to completely bypass channel stats but keep all of the others? Another option might be to destroy stats when channels go away. I know in my environment, I care about consumer channel info and don't care about channel info for publishers. We have many more publishers than consumers, and as a generalization, our publishers are transient while consumers are long-lived with the same channel.

I'm sure there are other strategies that could be employed with regard to tuning channel stats in the management interface. I find the rate information very useful as a generalization and don't like the idea of turning off all rates data because of data we don't really need.

michaelklishin:
@gmr in other words, you find channel rate stats to be least useful? This can be done and maybe even disabled by default but we'd like to see what difference resolving #214 and #215 would make.

gmr:
Sure, sorry for the wall of text... Yeah, least useful and I believe the cause of the issue.

hairyhum:
Running with memory analysis showed that memory leak was caused by leak in ETS. Because `old_stats_fine_index` table wasn't cleared after channel closure it did overflow erlang VM binaries someway. 
I'm not sure why, but it looks like `ets` can consume binaries memory while storing something like `[{<0.28662.14>, {fine,{<0.28662.14>, {resource,<<"/">>,exchange,<<"amq.topic">>}}}}]`

gmr:
huh -- Do you think it'd leak on lists/strings? Might be an easy fix.  Memory leak aside, it still seems like a reasonable thing to turn of channel stats. For brokers/clusters with lots of channel churn, I see value in being able to turn them off. I've been poking through the code to see how big of a change that would be.

michaelklishin:
[...]

PR #217, opened 2016-06-06
Clean up old_stats_fine_index

hairyhum:
Related to #216 
When using `rates_mode` other than `none`, opening and closing multiple channels in single connection can cause memory leak to `binaries`.
This was due to weird `ets` behaviour caused by growing `old_stats_fine_index` table.
Detected with stress-testing tool, mentioned in #216


hairyhum:
@gmr can you please verify if it helps?

gmr:
Initial results are very promising that it does indeed fix the memory leak. 

- Starting and stopping a large number of channels no longer leaves allocated memory
- Channel churn is no longer causing memory to grow in a 60 second test

I'm running a 10 minute test now and will report back.

gmr:
I'll give this a 👍 -- ran 10 minutes with settings that would previously caused the RabbitMQ container to eat 100% memory after 45 seconds.

screen shot 2016-06-06 at 11 45 00 am

PR #227, opened 2016-06-13
Partially re-enables `rabbitmqadmin` HTTPS support for Python < 2.7.9

### LINKED ISSUE #225: Python 2.6 doesn't provide 'ssl.create_default_context'
Running `rabbitmqadmin` on Python 2.6 fails with an error "AttributeError: 'module' object has no attribute 'create_default_context'". This is broken on master and on 3.6.2, but works on 3.6.1.

```
  File "/usr/local/bin/rabbitmqadmin", line 991, in <module>
    main()
  File "/usr/local/bin/rabbitmqadmin", line 413, in main
    method()
  File "/usr/local/bin/rabbitmqadmin", line 607, in invoke_declare
    self.put(uri, json.dumps(upload))
  File "/usr/local/bin/rabbitmqadmin", line 439, in put
    return self.http("PUT", "%s/api%s" % (self.options.path_prefix, path), body)
  File "/usr/local/bin/rabbitmqadmin", line 449, in http
    ssl_ctx = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
AttributeError: 'module' object has no attribute 'create_default_context'
```

According to [the rabbitmqadmin documentation](https://www.rabbitmq.com/management-cli.html) and the version check at the top of the `rabbitmqadmin` file, Python 2.6 is the minimum version of Python supported, but [ssl.create_default_context was added in 2.7.9](https://docs.python.org/2/library/ssl.html).

I'm not sure if this is a documentation issue (in which case the HTML page and the version check would need to be updated) or a bug in `rabbitmqadmin`'s compatibility introduced in #149.

michaelklishin:
Fixes #225 (see commit and issue comments: there is no low risk way to really make this work in `3.6.x` release series with modern RabbitMQ versions).

I'm going to update the docs to list 2.7.9 as the minimum required version for those using HTTPS.

### DIFF
--- a/bin/rabbitmqadmin
+++ b/bin/rabbitmqadmin
@@ -17,7 +17,7 @@
 
 import sys
 if sys.version_info[0] < 2 or (sys.version_info[0] == 2 and sys.version_info[1] < 6):
-    print("Sorry, rabbitmqadmin requires at least Python 2.6.")
+    print("Sorry, rabbitmqadmin requires at least Python 2.6 (2.7.9 when HTTPS is enabled).")
     sys.exit(1)
 
 from optparse import OptionParser, TitledHelpFormatter
--- a/bin/rabbitmqadmin
+++ b/bin/rabbitmqadmin
@@ -444,19 +444,34 @@ class Management:
     def delete(self, path):
         return self.http("DELETE", "%s/api%s" % (self.options.path_prefix, path), "")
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
[rabbitmqadmin docs](https://www.rabbitmq.com/management-cli.html) are updated.

michaelklishin:
Will re-submit against `stable`.

PR #229, opened 2016-06-14
Display GC statistic on node page

### LINKED ISSUE #208: Include GC and context switches stats into node information page
`erlang:statistics(garbage_collection)` doesn't provide much info but it's already something. `erlang:statistics(context_switches)` would also be useful. We should report those per unit of time, not total since the time the node was started.

hairyhum:
Fixes #208


michaelklishin:
This works as expected but could use a chart and more "standard" rate reporting (e.g. key named gc_num_details.rate), so still needs work.

Issue #211, opened 2016-05-25
Include reductions and GC stats on individual connection pages


michaelklishin:
As returned by `erlang:process_info/{1,2}`.

Issue #210, opened 2016-05-25
Include reductions and GC stats on individual channel pages


michaelklishin:
As returned by `erlang:process_info/{1,2}`.

Issue #209, opened 2016-05-25
Include reductions and GC stats on individual queue pages


michaelklishin:
As returned by `erlang:process_info/{1,2}`.

PR #232, opened 2016-06-16
Display GC and reductions on connection page

hairyhum:
Part of rabbitmq/rabbitmq-management#211

### DIFF
--- a/priv/www/js/tmpl/connection.ejs
+++ b/priv/www/js/tmpl/connection.ejs
@@ -68,6 +68,7 @@
  <td><%= connection.channel_max %> channels</td>
 </tr>
 </table>
+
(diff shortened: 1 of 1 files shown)


hairyhum:
Added reductions and GC to the channel page.
https://github.com/rabbitmq/rabbitmq-management/issues/210

michaelklishin:
This works as expected but could use a chart and more "standard" rate reporting (e.g. key named gc_num_details.rate), so still needs work.

dcorbacho:
Solves #209, #210, #211 together with https://github.com/rabbitmq/rabbitmq-common/pull/107 and https://github.com/rabbitmq/rabbitmq-server/pull/858

PR #248, opened 2016-07-13
Add help messages for VM metrics

### LINKED ISSUE #247: No descriptions for runtime metric charts
In the mgmt web UI, there are some charts(overview -> more about this node -> advanced)
`GC operations, GC bytes reclaimed, Context switch operations`
There are no descriptions for these chart

Also see:[rabbitmq-users](https://groups.google.com/forum/#!topic/rabbitmq-users/7OH1DCx33M8)

dcorbacho:
Fixes #247.

### DIFF
--- a/priv/www/js/help.js
+++ b/priv/www/js/help.js
@@ -328,6 +328,18 @@ HELP = {
         <dd>Rate at which queue index segment files are written. </dd>\
       </dl>',
 
+    'gc-operations':
+    'Number of garbage collection operations since the system started.',
+
+    'gc-bytes':
+    'Number of bytes reclaimed by the garbage collector since the system started.',
+
+    'context-switches-operations':
+    'Total number of context switches since the system started.',
+
+    'process-reductions':
+    'Number of reductions executed by the process.',
+
     'foo': 'foo' // No comma.
 };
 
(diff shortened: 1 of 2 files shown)


priv/www/js/help.js
```diff
@@ -328,6 +328,18 @@ HELP = {
         <dd>Rate at which queue index segment files are written. </dd>\
       </dl>',
 
+    'gc-operations':
+    'Number of garbage collection operations since the system started.',
```
michaelklishin: Is it since the system was started, though? We report it for the last N seconds.

priv/www/js/help.js
```diff
@@ -328,6 +328,18 @@ HELP = {
...
+
+    'gc-bytes':
+    'Number of bytes reclaimed by the garbage collector since the system started.',
```
michaelklishin: Same as above.

priv/www/js/help.js
```diff
@@ -328,6 +328,18 @@ HELP = {
...
+
+    'context-switches-operations':
+    'Total number of context switches since the system started.',
```
michaelklishin: Same as above.

PR #252, opened 2016-07-14
Make setup_extensions more resilient

### LINKED ISSUE #251: Management plugin without web_ui javascript
I have plugin that just mounts new api endpoint without touching UI.

```erlang
dispatcher() -> [{["metrics"],        rabbit_mgmt_metrics_handler, []}].
web_ui()     -> [].
```

this apparently breaks web ui because `/api/extensions` renders this:

```javascript
[[],{"javascript":"dispatcher.js"}]
```
[`setup_extensions`](
https://github.com/rabbitmq/rabbitmq-management/blob/5e1e841a936d6a48a3a2f3c49d41d7d59fe91ba8/priv/www/js/main.js#L140) always expects javascript key.

Question: what should `web_ui` return in this case? or maybe `setup_extensions` could be patched?

Thank you.

deadtrickster:
Fixes #251
Allows api-only management plugins.

### DIFF
--- a/priv/www/js/main.js
+++ b/priv/www/js/main.js
@@ -139,10 +139,13 @@ function update_vhosts() {
 
 function setup_extensions() {
     var extensions = JSON.parse(sync_get('/extensions'));
-    extension_count = extensions.length;
+    extension_count = 0;
     for (var i in extensions) {
         var extension = extensions[i];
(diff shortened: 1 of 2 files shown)


deadtrickster:
sorry wrong branch

Issue #258, opened 2016-07-20
Unsupported media type 415 response code


Gsantomaggio:
Executing `hop` tests on master I got the following error in different contexts:

```
./gradlew check -i
....
....
com.rabbitmq.http.client.ClientSpec > PUT /api/cluster-name FAILED
    org.springframework.web.client.HttpClientErrorException: 415 Unsupported Media Type
        at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:91)
        at org.springframework.web.client.RestTemplate.handleResponseError(RestTemplate.java:615)
        at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:573)
        at org.springframework.web.client.RestTemplate.execute(RestTemplate.java:544)
        at org.springframework.web.client.RestTemplate.put(RestTemplate.java:393)
        at com.rabbitmq.http.client.Client.setClusterName(Client.java:456)
```

It is the same issue fixed here: https://github.com/rabbitmq/rabbitmq-management/issues/107

should we copy this fix:

```
content_types_accepted(ReqData, Context) ->
   {[{'*', accept_content}], ReqData, Context}.
```
to the other contexts ?

michaelklishin:
@Gsantomaggio sounds good.

PR #259, opened 2016-07-20
Accept any content-type

Gsantomaggio:
Fixes https://github.com/rabbitmq/rabbitmq-management/issues/258 

Tested using `hole`
```
➜  hole git:(master) ✗ go test -v
=== RUN   TestRabbitHole
Running Suite: Rabbithole Suite
===============================
Random Seed: 1469026585
Will run 52 of 52 specs

••••••••••••••••••••••••••••••••••••••••••••••••••••
Ran 52 of 52 Specs in 13.169 seconds
SUCCESS! -- 52 Passed | 0 Failed | 0 Pending | 0 Skipped --- PASS: TestRabbitHole (13.17s)
PASS
```

`hop` tests passed for `Unsupported media type 415 response code` but there are still problems when the tests try to access to a deleted resource:

```
om.rabbitmq.http.client.ClientSpec > PUT /api/permissions/{vhost}/:user when vhost DOES NOT exist FAILED
    org.springframework.web.client.ResourceAccessException: I/O error on PUT request for "http://127.0.0.1:15672/api/users/hop-user1":null; nested exception is org.apache.http.client.ClientProtocolException
        at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:584)
        
        ...when vhost DOES NOT exist(ClientSpec.groovy:990)
```

I think because it is raised a `ResourceAccessException`  and [here](https://github.com/rabbitmq/hop/blob/master/src/main/java/com/rabbitmq/http/client/Client.java#L532)  it is not handled. (Could be ?)

### DIFF
--- a/src/rabbit_mgmt_wm_bindings.erl
+++ b/src/rabbit_mgmt_wm_bindings.erl
@@ -54,7 +54,7 @@ resource_exists(ReqData, {Mode, Context}) ->
     end.
 
 content_types_accepted(ReqData, Context) ->
-   {[{<<"application/json">>, accept_content}], ReqData, Context}.
+    {[{'*', accept_content}], ReqData, Context}.
 
 allowed_methods(ReqData, {Mode, Context}) ->
     {case Mode of
--- a/src/rabbit_mgmt_wm_cluster_name.erl
+++ b/src/rabbit_mgmt_wm_cluster_name.erl
@@ -38,7 +38,7 @@ content_types_provided(ReqData, Context) ->
    {[{<<"application/json">>, to_json}], ReqData, Context}.
 
 content_types_accepted(ReqData, Context) ->
-   {[{<<"application/json">>, accept_content}], ReqData, Context}.
+   {[{'*', accept_content}], ReqData, Context}.
 
 allowed_methods(ReqData, Context) ->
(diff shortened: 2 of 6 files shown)


michaelklishin:
@Gsantomaggio if `ResourceAccessException` can be boiled down to a "not found" error, then sure, let's handle it. Feel free to test against the [Ruby HTTP API client](https://github.com/ruby-amqp/rabbitmq_http_api_client) and Langohr as well ;)

Gsantomaggio:
@michaelklishin Ok perfect, let me test it using Ruby and Clojure

Gsantomaggio:
Ruby tests OK:
```
Finished in 6 minutes 41 seconds (files took 0.59771 seconds to load)
1074 examples, 0 failures, 4 pending
```

**Even if this is not related to this brach** it is worth to mention that this test  fails [here](https://github.com/ruby-amqp/rabbitmq_http_api_client/blob/master/spec/integration/client_spec.rb#L554) because we changed the [`HTTP get`](https://github.com/rabbitmq/rabbitmq-management/issues/68) .

Using `ack_requeue_false` fixes the problem:
```
xs = subject.get_messages("/", q.name, :count => 10, :ackmode => "ack_requeue_false", :encoding => "auto")
```

PR #261, opened 2016-07-21
Translate rabbitmqadmin tests from python to erlang

### LINKED ISSUE #234: Restore rabbitmqadmin test suite
As part of #218 we now run tests on nodes with non-standard ports. `rabbitmqadmin` test suite therefore cannot run. It needs to be adapted and included into `make tests` if possible.

kjnilsson:
Fixes #234


michaelklishin:
When merging into master we need to port the test changes in https://github.com/rabbitmq/rabbitmq-management/commit/ed7c490234f5e9f890219360b508d18354d0eda5 as well.

PR #267, opened 2016-08-01
Fix aggregation of messages and queue stats with unnamed tables

### LINKED ISSUE #266: Rates on overview tab are always 0
Originally reported as #265 without any details.

To reproduce, start a PerfTest instance and observe traffic flow and messages rates on the
connection(s), channel(s), and queue(s) pages but not Overview.

The only relevant change in 3.6.4 seems to be https://github.com/rabbitmq/rabbitmq-management/pull/257, which resolved #250.

dcorbacho:
Fixes #266

### DIFF
--- a/src/rabbit_mgmt_stats.erl
+++ b/src/rabbit_mgmt_stats.erl
@@ -146,10 +146,10 @@ sum([{T1, Id} | StatsN]) ->
                   end, AllIds),
     sum(StatsN, T).
 
-sum(StatsN, {_Table, IndexTable, _KeyIndexTable} = T) ->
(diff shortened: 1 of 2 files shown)


michaelklishin:
I think branch name here is supposed to be `rabbitmq-management-266` but no worries.

michaelklishin:
With this I get message rates reported as expected but only some of the time. I couldn't pin point what affects the outcome yet. This happens both with `PerfTest` and a Bunny REPL session that publishes over 2 connections to 2 vhosts in a loop.

michaelklishin:
No errors in either log file, too.

PR #287, opened 2016-09-19
Allow providing path when using #/login

snepomnyachi:
In certain cases we would like to allow our users to navigate to specific view after logging in.
e.g.: 
`/#/login/user/password/exchanges/rmq-stresstest/amq.direct`

We add new rule for login GET that navigates to the tail of the url 
after successful login.


michaelklishin:
Please base your PR on the `stable` branch, too.

PR #290, opened 2016-09-29
VHost limits page

### LINKED ISSUE #289: List configured VHost limits
There are new VHost parameters to limit connection and queue count on VHost that should be listed in management API.

hairyhum:
Fixes #289 

Limits page displays limit parameters.


hairyhum:
Added limits and operator policies management.
Added several new API requests.

src/rabbit_mgmt_wm_operator_policies.erl
```diff
@@ -48,7 +48,7 @@ to_json(ReqData, Context) ->
       ["priority"], ReqData, Context).
 
 is_authorized(ReqData, Context) ->
-    rabbit_mgmt_util:is_authorized_admin(ReqData, Context).
```
michaelklishin: Operator policies are supposed to be managed by administrators only.

hairyhum: This handler doesn't serve `PUT` or `DELETE` requests, only `GET`.
We must show operator policies to management users, otherwise it would be confusing.

PR #316, opened 2016-12-27
Add topic permissions endpoints

acogoluegnes:
References rabbitmq/rabbitmq-server#505


src/rabbit_mgmt_dispatcher.erl
```diff
@@ -110,16 +110,20 @@ dispatcher() ->
      {"/vhosts",                                               rabbit_mgmt_wm_vhosts, []},
      {"/vhosts/:vhost",                                        rabbit_mgmt_wm_vhost, []},
      {"/vhosts/:vhost/permissions",                            rabbit_mgmt_wm_permissions_vhost, []},
+     {"/vhosts/:vhost/topic-permissions",                      rabbit_mgmt_wm_topic_permissions_vhost, []},
```
michaelklishin: Please use underscores, as we already do, e.g. `operator_policies`.

src/rabbit_mgmt_dispatcher.erl
```diff
@@ -110,16 +110,20 @@ dispatcher() ->
...
      %% /connections/:connection is already taken, we cannot use our standard scheme here
      {"/vhosts/:vhost/connections",                            rabbit_mgmt_wm_connections_vhost, []},
      %% /channels/:channel is already taken, we cannot use our standard scheme here
      {"/vhosts/:vhost/channels",                               rabbit_mgmt_wm_channels_vhost, []},
      {"/users",                                                rabbit_mgmt_wm_users, []},
      {"/users/:user",                                          rabbit_mgmt_wm_user, []},
      {"/users/:user/permissions",                              rabbit_mgmt_wm_permissions_user, []},
+     {"/users/:user/topic-permissions",                        rabbit_mgmt_wm_topic_permissions_user, []},
```
michaelklishin: Same as above.

src/rabbit_mgmt_wm_definitions.erl
```diff
@@ -257,6 +259,7 @@ rw_state() ->
     [{users,              [name, password_hash, hashing_algorithm, tags]},
      {vhosts,             [name]},
      {permissions,        [user, vhost, configure, write, read]},
+     {topic_permissions,  [user, vhost, name, pattern]},
```
michaelklishin: I think `exchange` would be clearer than `name`.

src/rabbit_mgmt_dispatcher.erl
acogoluegnes: We also use dashes (e.g. `aliveness-test`, `cluster-name`, and also deprecated `all-configuration`), so I used them for `topic-permissions` and also recently for `global-parameters`. I don't mind changing those 2, but we won't be consistent anyway.
`operator_policies` and `vhost_limits` are in 3.7.0, so maybe we can still change them?

michaelklishin: OK, we will change the new endpoints in master before it's too late then.

PR #356, opened 2017-03-03
Make HTTP API stats.html link relative

binarin:
Otherwise this link is broken on the page
https://raw.githack.com/rabbitmq/rabbitmq-management/rabbitmq_v3_6_6/priv/www/api/index.html
(currently mentioned at http://www.rabbitmq.com/management.html )

I've tested it, relative link also works when the doc is show by management plugin in a live rabbit instance.


michaelklishin:
I believe we've tried this before and there was something subtle that broke. I will give it a try and cherry-pick to `live`. Thank you.

PR #367, opened 2017-03-23
Add support for non-JSON encoding of replies

binarin:
One of the low-hanging fruits in HTTP API optimization.
E.g. serializing information about 10000 queues:
- For JSON it takes 2 seconds on my machine and more than 1GB of RAM
- For BERT it takes 0.2 seconds and a negligible amount of RAM

If the approach in general is OK, I'm going to replace all other
instances of `<<"application/json">>` everywhere.

### DIFF
--- a/src/rabbit_mgmt_util.erl
+++ b/src/rabbit_mgmt_util.erl
@@ -28,7 +28,8 @@
 -export([with_decode/4, not_found/3, amqp_request/4]).
 -export([with_channel/4, with_channel/5]).
 -export([props_to_method/2, props_to_method/4]).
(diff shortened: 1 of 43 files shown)


michaelklishin:
Thank you! I personally don't see any issues with this approach.

kjnilsson:
@binarin thank you - the approach looks generally fine to me.

dcorbacho:
@binarin thanks, looks good to me too!

michaelklishin:
@binarin so can you please get this PR into a state that's ready for QA and merging?

binarin:
Done

PR #369, opened 2017-03-24
Add ability to skip sorting of results

binarin:
Another low-hanging fruit for HTTP API optimization.

Allow user to specify empty sort parameter to avoid paying the sorting
cost. As for some applications (like monitoring) it doesn't matter
whether the result is sorted or not. And e.g. for 10000 queues
performing the sort using the default sorting order costs ~0.3 second
on my machine.

### DIFF
--- a/src/rabbit_mgmt_util.erl
+++ b/src/rabbit_mgmt_util.erl
@@ -255,6 +255,11 @@ reply_list_or_paginate(Facts, ReqData, Context) ->
 sort_list(Facts, Sorts) -> sort_list(Facts, Sorts, undefined, false,
   undefined).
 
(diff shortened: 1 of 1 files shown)


kjnilsson:
@binarin not sorting would make the paging behaviour undefined but as this is only for an explicitly empty sorting parameter I think that is ok. I'm guessing this would typically be used without paging?

binarin:
Yes, it's primarily intended for monitoring, where we want to scrape everything in a single request.

kjnilsson:
Ok looks fine to me. Any concerns @dcorbacho ?

dcorbacho:
No, I think it is good.

PR #374, opened 2017-04-10
Simpler way to create amqp methods in binding deletion.

### LINKED ISSUE #243: UI operation for binding deletion doesn't respect extra binding arguments
See [this mailing list thread](https://groups.google.com/forum/#!topic/rabbitmq-users/m4oaXLfzlT0).

hairyhum:
Fixes #243 
Formatting/parsing a binding loses information about argument types,
so unbinding can fail if arguments are not of "standard" type.
For example if number argument is not `long`

### DIFF
--- a/src/rabbit_mgmt_wm_binding.erl
+++ b/src/rabbit_mgmt_wm_binding.erl
@@ -25,7 +25,6 @@
 -include_lib("amqp_client/include/amqp_client.hrl").
 
 %%--------------------------------------------------------------------
-
 init(_, _, _) -> {upgrade, protocol, cowboy_rest}.
 
 rest_init(Req, _Config) ->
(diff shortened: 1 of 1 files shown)


michaelklishin:
@hairyhum can you recommend a way to reproduce this?

hairyhum:
From a client which supports argument typing:
* create a binding with a numerical `long` argument
* create a binding with a numerical `signedint` (or other non-long) argument.

The first one will be deletable in 3.6.9, deleting the second one will return 204, but the binding will persist.

michaelklishin:
Reproduced this with the Go client and the following bindings:

``` go
	if err = channel.QueueBind(
		"management243-q1", // name of the queue
		"q1",        // bindingKey
		*exchangeName,   // sourceExchange
		false,      // noWait
	        amqp.Table{
		"int16": int16(100),
		"int32": int32(100),
		"int64": int64(100),
		},        // arguments
	); err != nil {
		log.Fatalf("Queue Bind: %s", err)
	}
	if err = channel.QueueBind(
		"management243-q2", // name of the queue
		"q2",        // bindingKey
		*exchangeName,   // sourceExchange
		false,      // noWait
		amqp.Table{
		"int8": int8(2),
		},        // arguments
	); err != nil {
		log.Fatalf("Queue Bind: %s", err)
	}
```

PR #391, opened 2017-05-04
Ensure cowboy Req is returned from `with_decode`

kjnilsson:
Reversing initial fix in 74e9e36cf7a8baa9e5a0d6714fb7111dee5a37f8

Usages of `with_decode` closed over the ReqData and returned the
instance before the body decode. This had the effect of causing an error
after a  POST or PUT request had been issues on the same keep-alive
connection.


michaelklishin:
Wow, good find.

PR #399, opened 2017-05-26
Augment queues after sorting (if possible)

binarin:
Profiling shows that most expensive part of listing
queues/exchanges/channels etc. is calculating various rates using
exometer functionality.

Given that Web UI uses pagination and the default sort order (and some
other reasonable sort orders) actually doesn't require augmented data
to perform sorting and paging, sometimes we can postpone augmentation
until the data set is a lot smaller.

This patch lays basic groundwork for this approach and also implements
such a behavior for queues.

Test on my workstation shows that on a broker with 10000 warmed-up queues
reply time for a 100-items page with default sort orders drops from 6
seconds to 0.3 seconds. These are timings when
rabbit_mgmt_db:submit_cached/4 doesn't hit, and caching slightly
improves the situation. But this patch strives to get rid of this
overhead completely.

### DIFF
--- a/src/rabbit_mgmt_util.erl
+++ b/src/rabbit_mgmt_util.erl
@@ -315,8 +315,9 @@ unaugmented_reply_list_or_paginate(BasicFacts,
       ReqData, Context,
       fun(Pagination) ->
               DefaultSorts = ["vhost", "name"],
-              SortParam = merge_sorts_from_req(DefaultSorts, ReqData),
-              AugmentAfterPagination = is_pagination_requested(Pagination) andalso CanAugmentAfterPaginationPredicate(SortParam),
+              SortParam = get_value_param(<<"sort">>, ReqData),
+              MentionedSortColumns = lists:usort(merge_sorts(DefaultSorts, SortParam)),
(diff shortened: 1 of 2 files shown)


kjnilsson:
Looks like sensible approach to me, thank you! 

The only potential draw-back is that there will be a substantial performance difference once someone decide to use sorting.

binarin:
I've fixed it.

michaelklishin:
Both management test suites and test suites of 4 HTTP API clients pass with this => LGTM.

binarin:
I'm not sure that I'll have to implement this for other things, such as exchanges or channels. Should I file an issue?

michaelklishin:
@binarin it's worth exploring for connections and channels. A large number of exchanges is something pretty rare to come by.

carlhoerberg:
Edge case, but our shared clusters we have a lot of them, as each vhost as at least 7 exchanges, so 7000 exchanges is not rare for us..

michaelklishin:
@carlhoerberg fair enough. If we can do the same for exchanges, I'm all for it, of course.

michaelklishin:
#417 extends this approach to exchanges. Connections and channels will require a slightly modified version but we hope the same idea is applicable.

PR #415, opened 2017-06-13
Use rabbit_mgmt_format:args_hash to get args hash in bindings API and tests

hairyhum:
Part of fix for rabbitmq/rabbitmq-server#1243

Requires https://github.com/rabbitmq/rabbitmq-management-agent/pull/47


michaelklishin:
Should this be backported to `stable`?

michaelklishin:
That's a good point. OK.

PR #418, opened 2017-06-16
Extend resource augmentation optimisation.

### LINKED ISSUE #417: extend /queues optimisation approach
Extend the approach in #399 to sequence the `sort`, `page` and `augment` operations in the most optimal way for a given query and also to cover exchanges and vhosts as well as queues.

The idea is to first analyse the request and translate it into a strategy that will execute the most optimal sequence of augmentation and paging operations. Take this pseudo pattern match expression where the first item is a bool indicating whether paging is requested, the second whether the sorting columns include "extended" properties, i.e. properties not present in the resource record and the third item whether any extended columns are requested. Here we return a pipeline of operations to execute to augment the request. Not all include the expensive cluster-wide `augment` operation. Also there is a case where we can page _before_ we augment which will also yield efficiency gains.

```
{false, basic, basic}
    [sort];
{false, extended, _}
    [augment, sort];
{true, basic | default, basic}
    [sort, page]
{true, extended, _}
    [augment, sort, page];
{true, basic, extended}
    [sort, page, augment]

```

kjnilsson:
Analyses the http request paging, column and sorting parameters and builds
up the most optimal augmentation pipeline. This allows us to skip
cluster-wide augmentation when only basic fields are requested as well
as reduce the augmentation set in certains scenarios prior to
augmentation.

`hop` and `rabbit-hole` tests suites pass.

Fixes #417

### DIFF
--- a/src/rabbit_mgmt_util.erl
+++ b/src/rabbit_mgmt_util.erl
@@ -34,13 +34,16 @@
 -export([with_decode/5, decode/1, decode/2, set_resp_header/3,
          args/1]).
 -export([reply_list/3, reply_list/5, reply_list/4,
-         sort_list/2, destination_type/1, reply_list_or_paginate/3,
-         unaugmented_reply_list_or_paginate/5]).
+         sort_list/2, destination_type/1, reply_list_or_paginate/3
+         ]).
 -export([post_respond/1, columns/1, is_monitor/1]).
 -export([list_visible_vhosts/1, b64decode_or_throw/1, no_range/0, range/1,
          range_ceil/1, floor/2, ceil/1, ceil/2]).
--export([pagination_params/1, maybe_filter_by_keyword/4,
-         get_value_param/2]).
+-export([pagination_params/1,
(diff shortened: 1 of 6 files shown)


src/rabbit_mgmt_util.erl
```diff
@@ -355,6 +300,104 @@ merge_sorts(DefaultSorts, Extra) ->
...
+column_strategy(Cols, BasicColumns) ->
+    case Cols -- BasicColumns of
+        [] -> basic;
+        _ -> extended
+    end.
+
+% columns are [[binary()]] - this takes the first item
+columns_as_strings(all) -> all;
+columns_as_strings(Columns0) ->
+    Columns = lists:foldl(fun([C | _], Acc) ->
```
dcorbacho: Could use `lists:map` or lists comprehensions and avoid the `lists:reverse`

kjnilsson: makes sense.

PR #426, opened 2017-06-21
Fix rabbitmqadmin link on /cli page. Make its URI absolute so it alwa…

lukaswoj:
…ys works, even when /cli page gets accessed without trailing slash, like http://rabbitmq.server:15672/cli


michaelklishin:
Thank you. I will cherry-pick to `stable`.

PR #454, opened 2017-07-26
Change users bulk delete to use a deletion resource.

kjnilsson:
DELETE with a payload doesn't have prescribed semantics and at least
older version of the HTTP spec prescribe that the request uri should
full identify the resource to be deleted.

This change introduces a /users/bulk-delete resource to which we POST the
users that are to be deleted.

[#149484991]


dcorbacho:
I think it would be better to implement it on a different handler (i.e `rabbit_mgmt_wm_users_bulk_delete`, as in this way we allow to do a `GET` of `/users/bulk-delete` and a `POST` of `/users`.

kjnilsson:
Good point @dcorbacho

kjnilsson:
Ok I've updated it to use a new handler as suggested.

PR #469, opened 2017-08-14
CSS to mask passwords (uris) in UI by only showing on mouse hover

dcorbacho:
Part of https://github.com/rabbitmq/rabbitmq-federation-management/issues/15

### DIFF
--- a/priv/www/css/main.css
+++ b/priv/www/css/main.css
@@ -302,10 +302,10 @@ tr.alt2>td {
     border: 1px solid #e0e0e0;
 }
 
-#mouse_over {
+.mouse_over {
     opacity: 0;
 }
 
-#mouse_over:hover {
+.mouse_over:hover {
     opacity: 1;
 }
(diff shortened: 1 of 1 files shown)


priv/www/css/main.css
```diff
@@ -301,3 +301,11 @@ tr.alt2>td {
     background: -webkit-gradient(linear, left top, left bottom, color-stop(0, #f0f0f0),color-stop(1, #e0e0e0));
     border: 1px solid #e0e0e0;
 }
+
+#mouse_over {
```
acogoluegnes: Could it be a CSS class?

michaelklishin: That's what I recommended in the other two PRs :)

michaelklishin:
We decided to go with a different approach: mask credentials but display the URIs unconditionally. With display-on-hover it is impossible to see all links or shovels at once.

PR #470, opened 2017-08-14
Avoid attempts to grant dummy "internal" user permissions when loading definitions

michaelklishin:
Happens when definitions are loaded on node startup, so there is no
"current user" to grant permissions to.

Fixes rabbitmq/rabbitmq-server#1329.


michaelklishin:
This was reverted in https://github.com/rabbitmq/rabbitmq-management/commit/242b9bdd67467c90a7ed1402bdab3b3f7afcf430, it's insufficient.

PR #472, opened 2017-08-29
Introduce a function that redacts passwords in URIs

michaelklishin:
Part of https://github.com/rabbitmq/rabbitmq-federation-management/issues/15.

### DIFF
--- a/priv/www/js/formatters.js
+++ b/priv/www/js/formatters.js
@@ -570,7 +570,7 @@ function fmt_shortened_uri(uri) {
 function fmt_uri_with_credentials(uri) {
     if (typeof uri == 'string') {
         // mask password
-        var mask = /^([a-zA-Z0-9]+):\/\/(.*):(.*)@/;
+        var mask = /^([a-zA-Z0-9\+\-\.]+):\/\/(.*):(.*)@/;
         return uri.replace(mask, "$1://$2:[redacted]@");
     } else {
         return UNKNOWN_REPR;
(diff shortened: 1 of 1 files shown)


priv/www/js/formatters.js
```diff
@@ -567,6 +567,16 @@ function fmt_shortened_uri(uri) {
     }
 }
 
+function fmt_uri_with_credentials(uri) {
+    if (typeof uri == 'string') {
+        // mask password
+        var mask = /^([a-zA-Z0-9]+):\/\/(.*):(.*)@/;
```
lukebakken: The only URI schemes we'll run into at this point are `amqp` and `amqps`, but it appears that a couple other characters are valid in them: `-`, `+` and `.` ([RFC](http://www.ietf.org/rfc/rfc2396.txt)).

Something like this -

`^([a-zA-Z0-9-+.]+):\/\/([^:]*):([^@]*)@`

michaelklishin:
@lukebakken added. I'm working on another PR for federation-management.

priv/www/js/formatters.js
```diff
@@ -567,6 +567,16 @@ function fmt_shortened_uri(uri) {
     }
 }
 
+function fmt_uri_with_credentials(uri) {
+    if (typeof uri == 'string') {
+        // mask password
+        var mask = /^([a-zA-Z0-9\+\-\.]+):\/\/(.*):(.*)@/;
```
lukebakken: Sorry to be pedantic, but `\` isn't necessary to escape special chars when they appear in a character class (`[ ]`):

regex-without-backslashes

Issue #487, opened 2017-09-26
PUT followed by GET on exchange fails miserably


essen:
Placeholder ticket to fix an issue where the Req object is discarded, preventing from doing anything on the connection after a `PUT /api/exchanges/:vhost/:exchange`.

This issue only exists on master. Stable, and 3.6.x are working fine.

PR #488, opened 2017-09-26
Don't discard the new Req object after reading the body

essen:
This is a design flaw in Cowboy before Cowboy 2.0: discarding
the new variable prevents Cowboy from knowing the body has been
read, and as a result Cowboy attempts to read the body again when
there's nothing there (or there's a new request coming in instead,
which is considered to be the body).

#487

### DIFF
--- a/src/rabbit_mgmt_util.erl
+++ b/src/rabbit_mgmt_util.erl
@@ -737,7 +737,7 @@ get_node(Props) ->
 direct_request(MethodName, Transformers, Extra, ErrorMsg, ReqData,
                Context = #context{user = User}) ->
     with_vhost_and_props(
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@essen Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-management&pullRequestId=488)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-management&pullRequestId=488) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

PR #496, opened 2017-10-23
Update cowboy to 2.0

hairyhum:
New cowboy_req API.
New cowboy protocol API.
New static page handler.
Replace init/3 and rest_init/2 with init/2 in all handlers.

Requires https://github.com/rabbitmq/rabbitmq-web-dispatch/pull/30

[#151681379]


src/rabbit_mgmt_wm_static.erl
```diff
@@ -21,8 +21,7 @@
```
essen: I think this entire file is not necessary anymore. Cowboy 2.0's cowboy_static supports reading files from .ez archives thanks to @dumbbell's patch. See https://git.ninenines.eu/cowboy.git/commit/?id=2166733628bbab0eb77eeed58bdf204727c48ec6

hairyhum: This module uses this feature. Most of the code handles routing logic for management plugin applications. Is there a better way to do that?

essen: Nevermind I was reading the code too fast.

PR #542, opened 2018-01-16
Make requeue the default option in the Get Messages section to avoid accidentally deleting messages

dparkinson:
## Proposed Changes

Change the order of the Ack Mode drop down under the Get Messages section of a Queue so that requeue is the first / default option. 

The proposed change is so that users do not accidentally remove messages from the queue without confirming that they should be permanently deleted. This is only a UI change and does not affect any logic of the Get operation.

## Types of Changes

- [X] Cosmetics (whitespace, appearance)
- [X] Usability (improved user interaction)

## Checklist

- [X] I have read the `CONTRIBUTING.md` document
- [X] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [X] All tests pass locally with my changes

## Further Comments

Before:
requeue-default-before

After: 
requeue-default-after

### DIFF
--- a/priv/www/js/tmpl/queue.ejs
+++ b/priv/www/js/tmpl/queue.ejs
@@ -226,10 +226,10 @@
           <th><label>Ack Mode:</label></th>
           <td>
             <select name="ackmode">
-                <option value="ack_requeue_false">Ack message requeue false</option>
                 <option value="ack_requeue_true">Nack message requeue true</option>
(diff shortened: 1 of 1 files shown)


michaelklishin:
Our team agrees this is a good idea. I am QA'ing it.

michaelklishin:
Thank you!

PR #550, opened 2018-02-06
Remove unnecessary call to xmerl_ucs:from_utf8

essen:
This commit removes a call to xmerl_ucs:from_utf8 which was used
for two purposes: converting the binary payload into an Erlang
Unicode string (a list of integer code points), and validating
that the payload is indeed utf8.

The call was necessary when it was introduced because mochijson
only supported Unicode strings. Nowadays JSON libraries also
support utf8 binaries so converting is not necessary.

The function also starts by doing a binary_to_list/1 call which
creates a lot of garbage and can lead to OOM situations when the
queue is large. We ended up with the payload binary, the temporary
list and the final Unicode string in memory. With this patch we
only ever have the binary in memory.

To validate that the payload is utf8 and keep the functionality
intact a small function was added that makes use of the /utf8
binary matching specifier.

Since this was the only xmerl function used in the management
plugin I have also removed xmerl from the LOCAL_DEPS.

I have added a test for the different encodings that can be
requested to make sure that nothing was broken when doing
the change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ endef
 
 DEPS = rabbit_common rabbit amqp_client cowboy cowlib rabbitmq_web_dispatch rabbitmq_management_agent
 TEST_DEPS = rabbitmq_ct_helpers rabbitmq_ct_client_helpers proper
-LOCAL_DEPS += xmerl mnesia ranch ssl crypto public_key
+LOCAL_DEPS += mnesia ranch ssl crypto public_key
 
 # FIXME: Add Ranch as a BUILD_DEPS to be sure the correct version is picked.
 # See rabbitmq-components.mk.
--- a/src/rabbit_mgmt_wm_queue_get.erl
+++ b/src/rabbit_mgmt_wm_queue_get.erl
@@ -162,13 +162,14 @@ maybe_truncate(Payload, Len) ->
 
 payload_part(Payload, Enc) ->
     {PL, E} = case Enc of
-                  auto -> try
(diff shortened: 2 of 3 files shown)


essen:
Up to you whether this should go in 3.6.x too. Cheers!

michaelklishin:
No, it probably should not. I'm QA'ing this right now, thank you!

gerhard:
**No action required, adding for context**

This was discovered while investigating RabbitMQ 3.6.14 node crashes when a queue with 5GB of messages was being downloaded through the Management API, via `POST /api/queues/VHOST/QUEUE/get`.

The process with an initial call of `cowboy_protocol:init/4` was `Garbing...` and using 18.9GB of memory. Erlang VM 19.3.6.4 crashed due to `eheap_alloc: Cannot allocate 29247236448 bytes of memory (of type "old_heap")`. The VM had already allocated 22GB of memory:

```
total: 24457794736
processes: 21850326288
processes_used: 21845621216
system: 2607468448
atom: 1041593
atom_used: 1038406
binary: 2457997680
code: 25834937
ets: 91379304
```

[**More context** - RabbitMQ team internal only](https://discuss.zendesk.com/agent/tickets/75578)

michaelklishin:
Hm, then I guess it should go into 3.6.x.

michaelklishin:
Backported to 3.7.x.

michaelklishin:
Backported to 3.6.x.

Issue #575, opened 2018-05-23
Virtual hosts imported from a definitions file need a moment to initialise


michaelklishin:
See [this rabbitmq-users thread](https://groups.google.com/forum/#!topic/rabbitmq-users/BQpbd6KP-VQ). As of 3.7.0 virtual hosts need a bit of time to initialise their process tree.

It is also possible that a vhost is initialised on the local node but not on others yet some schema definition operations (e.g. queue declarations) are applied on different nodes, e.g. due to queue master locators effects.

michaelklishin:
Updating the milestone because

 * The management plugin part of the fix was *NOT* cherry-picked for `3.7.6` 🤦‍♂️.
 * #591 introduces a higher timeout.

So having the milestone say `3.7.6` would be too confusing.

PR #591, opened 2018-07-27
Bumping vhost start timeout

kitsirota:
This issue is new since we upgraded from 3.6.x.  We run 5 node clusters running RMQ 3.7.7/Erlang 20.3.8.1.  When we reach about 15 vhosts, new vhosts can take longer than 15s to create.  This typically results in unhealthy vhosts with 1+ "stopped" nodes.  

## Proposed Changes

We would like to bump the limit to 45 seconds to mitigate having to detect failed nodes with an external monitoring solution and start them via the /api/vhosts/name/start/node endpoint.

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #575)

## Checklist

- [x ] I have read the `CONTRIBUTING.md` document
- [x ] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)

### DIFF
--- a/src/rabbit_mgmt_wm_vhost.erl
+++ b/src/rabbit_mgmt_wm_vhost.erl
@@ -95,9 +95,9 @@ put_vhost(Name, Trace, Username) ->
     Result = case rabbit_vhost:exists(Name) of
         true  -> ok;
         false -> rabbit_vhost:add(Name, Username),
-                 %% wait for up to 15 seconds for the vhost to initialise
+                 %% wait for up to 45 seconds for the vhost to initialise
                  %% on all nodes
-                 case rabbit_vhost:await_running_on_all_nodes(Name, 15000) of
+                 case rabbit_vhost:await_running_on_all_nodes(Name, 45000) of
                      ok               ->
                          maybe_grant_full_permissions(Name, Username);
                      {error, timeout} ->


michaelklishin:
This is merely a workaround for a problem somewhere else. 15 seconds to create a virtual host is pretty extreme.

michaelklishin:
I am not against merging this but please post some details on how it can be reproduced (and give 3.7.8-rc.1 a shot, it has non-trivial optimizations in virtual host recovery) to the mailing list.

kitsirota:
@michaelklishin I absolutely agree, this is purely a bandaid for a condition that should probably be handled asynchronously.    We're having trouble identifying the root cause of the variable request times.

In our use-case, we're deploying containers in a Pivotal CloudFoundry deployment.  We seem to reach a point where vhosts take longer than 15 seconds to generate when we're up to about 300-400 containers spread across around 15 vhosts.  The underlying nodes dont seem to have any operational bottlenecks (no load/mem/io issues when this starts happening).

Thanks!

michaelklishin:
@kitsirota so, 300-400 application instances? How many connections do they open on average? (a ball park estimate would do)

michaelklishin:
So apparently the management part of #575 was not cherry-picked to `v3.7.x` 🤦‍♂️, so the bump per se may or may not be necessary but 45s is not an unreasonable value.

Issue #596, opened 2018-08-02
Expose `listener.cowboy_opts` to new style configuration file


michaelklishin:
Current Cowboy options can only be [configured via the classic config format](https://www.rabbitmq.com/web-dispatch.html). Much like with TLS options, the set of values
is known, doesn't change often, and can benefit from additional validation. In addition it would be easier to enforce defaults (e.g. enable HTTP response compression) with Cuttlefish.

It would be nice to support at least a subset of Cowboy options in the new style format.

michaelklishin:
Filed here for visibility. This equally applies to Web STOMP, Web MQTT and rabbitmq-top.

asgs:
@michaelklishin  when will v3.7.8 be released?

michaelklishin:
2-3 weeks sounds most likely. Don't quote me on that.

michaelklishin:
It is [possible to configure Cowboy](http://www.rabbitmq.com/web-dispatch.html) using the classic format (so, `advanced.config`) today.

michaelklishin:
@asgs this is not a discussion forum. If you have a reliable way to reproduce, post it to the mailing list and we'll see if we can.

michaelklishin:
Reopening per some QA feedback from @acogoluegnes.

scm-git:
from the version 3.7.8, the default configure format is key-value,right? when I use the classic format, it reports wrong format when I start rabbitmq-server;
but the classic configure format works well at version 3.7.7.

michaelklishin:
@scm-git please [read the docs](http://www.rabbitmq.com/configure.html#configuration-files) and ask questions on the mailing list. Nothing has changed between 3.7.7 and 3.7.8. You can use either format.

PR #606, opened 2018-09-04
Make sure listener has a port even if it's not provided in the config

michaelklishin:
## Proposed Changes

This makes sure the listener always has a port specified, even if it's missing from the config file.

## Types of Changes

- [x] Bug fix (non-breaking change)

## Checklist

## Further Comments

Discovered while QA'ing #596, even though they are not strictly related.

[#159556318]


michaelklishin:
An example config file that would result in a node startup failure in 3.7.7 but not with this PR:

```ini
management.listener.server.shutdown_timeout = 7000
management.listener.server.max_keepalive = 120
management.listener.server.idle_timeout = 120
management.listener.server.inactivity_timeout = 120
management.listener.server.request_timeout = 120
management.listener.server.compress = true
```

PR #653, opened 2019-01-18
Add help and link to explain active consumer flag

acogoluegnes:
[#163298456]

References rabbitmq/rabbitmq-server#1838


priv/www/js/global.js
```diff
@@ -425,9 +425,15 @@ var HELP = {
 
     'filter-regex' :
     'Whether to enable regular expression matching. Both string literals \
-    and regular expressions are matched in a case-insensitive manner.<br/></br/> \
+    and regular expressions are matched in a case-insensitive manner.<br/><br/> \
     (<a href="https://developer.mozilla.org/en/docs/Web/JavaScript/Guide/Regular_Expressions" target="_blank">Regular expression reference</a>)',
 
+    'consumer-active' :
+    'Whether the consumer is active or not. Consumers are active by default. For a quorum queue, a consumer can be inactive \
```
hairyhum: Can we first mention single-active consumer scenario and then quorum queues?

priv/www/js/tmpl/consumers.ejs
```diff
@@ -31,7 +31,7 @@
         <td class="c"><%= fmt_boolean(consumer.ack_required) %></td>
         <td class="c"><%= fmt_boolean(consumer.exclusive) %></td>
         <td class="c"><%= consumer.prefetch_count %></td>
-        <td class="c"><%= fmt_boolean(consumer.single_active) %></td>
+        <td class="c"><%= fmt_boolean(consumer.active) %></td>
```
hairyhum: Can we also have the activity status here? Since we already have it somewhere.

priv/www/js/global.js
hairyhum: It would be nice to have some explanation of what does the word "active" mean. Something like "can get messages from the queue".

priv/www/js/tmpl/consumers.ejs
acogoluegnes: To be implementation specific, metrics fields are stored in Mnesia when some event is propagated, whereas fields in the CLI are not stored anywhere, they are computed on the fly. This is no big deal to add though, especially if it's useful in the channel page.

acogoluegnes:
Thanks @hairyhum for the feedback. This is ready for another round of QA.

PR #661, opened 2019-02-18
Set channel source in rabbit_mgmt_util rpc call to rabbit_channel

Ayanda-D:
Part of https://github.com/rabbitmq/rabbitmq-server/pull/1886


michaelklishin:
Backported to `v3.7.x`. @Ayanda-D please double check?

PR #663, opened 2019-03-01
Change stats merge function to concatenate strings

kjnilsson:
As else when consumers are spread across a cluster they are not all
returned when viewed from the queue page.

[#164302066]


michaelklishin:
Backported to `v3.7.x`.

PR #688, opened 2019-04-04
Catch failures when there is no user/vhost.

hairyhum:
## Proposed Changes

list_permissions operations may throw an exception if user/vhost
does not exist. Handlers use dirty_read to check for users/vhosts,
which is not enough if there is a concurrent deletion.
Return 404 instead of 500 in such cases.

To reproduce you need to start several applications, which will create users, list permissions and delete users.
List permission requests will sometimes fail with 500.
With this change they should fail with 404.

## Types of Changes

## Checklist


michaelklishin:
Backported to `v3.7.x`.

PR #693, opened 2019-04-23
Set msg rates in overview to global and basic in queue pages

dcorbacho:
I can't think of a reason why we won't show the last 8/24h for message rates when they're available - same as message counters - instead of default to `basic`.

Also removes warning from the console, when we can easily set the default for all the known rates.

[#165377789]

## Types of Changes

## Checklist


michaelklishin:
Backported to `v3.7.x`.

PR #694, opened 2019-04-23
Fix bytes prefix symbol

gerhard:
B is used for powers of 10, e.g. 10^6 bytes = 1MB
iB is used for powers of 2, e.g. 2^20 bytes = 1MiB

Since the division is using a power of 2, the symbol used also needs to
be a power of 2.

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue)

## Checklist

## Further Comments

This is somewhat related to rabbitmq/rabbitmq-server#1987

Before this change, 93048832 bytes were shown as 89MB. After this change, they are shown as 89MiB. This applies to all units:

image


michaelklishin:
Backported to `v3.7.x`.

PR #711, opened 2019-06-17
Definitions import timeout

hairyhum:
## Proposed Changes

When importing long definitions file, the request may fail silently due to `idle_timeout` in cowboy. [1](https://ninenines.eu/docs/en/cowboy/2.0/manual/cowboy_http/)
This timeout terminates the connection process silently without any logs.
The default timeout is 60 seconds and setting it in the configuration will impact all other requests, which is undesirable because the HTTP client can fail and there will be zombie processes. [2](https://github.com/ninenines/cowboy/issues/1278)

For the definitions import it's expected that it may take long time, therefore this PR notifies the cowboy stream process that the timeout should be disabled.
This is done after entire request body is received and should not lead to zombie processes, because the request does not wait for the client after that point.

Also added process reports, sent with HTTP 102 messages to prevent clients from timing out and closing the connection.

## Types of Changes

## Checklist

## Further Comments

Open questions:
- I'm not sure that infinity is the best value here, maybe just a large integer would be better.
- I didn't find and API to set the timeout from the request handler, the API I used is a bit hacky. @essen is there a better way to do that?

### DIFF
--- a/src/rabbit_mgmt_load_definitions.erl
+++ b/src/rabbit_mgmt_load_definitions.erl
@@ -74,4 +74,7 @@ load_definitions_from_file(File) ->
 
 load_definitions(Body) ->
     rabbit_mgmt_wm_definitions:apply_defs(
-      Body, ?INTERNAL_USER, fun () -> ok end, fun (E) -> {error, E} end).
+        Body, ?INTERNAL_USER,
+        fun () -> ok end,
+        fun (E) -> {error, E} end,
+        fun (_Msg) -> ok end).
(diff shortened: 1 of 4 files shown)


essen:
"This may be not the best way to configure this setting."

At the moment, it is. There will be a function added later on for configuring per-request options.

michaelklishin:
@essen thanks for confirming!

michaelklishin:
Backported to `v3.7.x`.

michaelklishin:
Backported to `v3.7.x`.

PR #717, opened 2019-07-16
Alternative copywriting for change log footer link

michaelklishin:
Per discussion with @gerhard.

### DIFF
--- a/priv/www/js/tmpl/layout.ejs
+++ b/priv/www/js/tmpl/layout.ejs
@@ -49,6 +49,6 @@
     <li><a href="https://www.rabbitmq.com/services.html" target="_blank">Commercial Support</a></li>
     <li><a href="https://www.rabbitmq.com/plugins.html" target="_blank">Plugins</a></li>
     <li><a href="https://www.rabbitmq.com/github.html" target="_blank">GitHub</a></li>
-    <li><a href="https://www.rabbitmq.com/changelog.html" target="_blank">Changelog</a></li>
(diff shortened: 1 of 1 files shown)


gerhard:
This is a step in the right direction. Would it be possible to add the rabbitmq_management version to this link? I'm thinking e.g.:

```diff
- <li><a href="https://www.rabbitmq.com/changelog.html" target="_blank">Latest Release</a></li>
+ <li><a href="https://www.rabbitmq.com/changelog.html?v=3.8.0" target="_blank">Latest Release</a></li>
```

michaelklishin:
My idea was to make the link more "catchy" without pinning to a particular version. What you're suggesting would make more sense for the version number link (not a part of this PR).

gerhard:
The version number would be tied to the rabbitmq_management that the user is accessing the link from.

For example, if I am running RabbitMQ v3.8.0-rc.1, the link in the Managemet UI would be `https://www.rabbitmq.com/changelog.html?v=3.8.0-rc.1`

The goal is to have a non-intrusive way of seeing which RabbitMQ version users are coming from when they check the latest RabbitMQ version.

michaelklishin:
That would require a different solution. This simply highlights the link that might have been not very prominent. Currently site analytics is not in a position to easily collect query arguments.

We agreed to look into a plugin that would collect such telemetry (and as all plugins, would be opt-in).
Note that products that build on top can choose to enable such plugin by default.

Issue #723, opened 2019-07-26
Unable to delete binding with undefined routing key


tomhreb:
Hi,
  
* RabbitMQ version: 3.7.15
* Erlang version: 22.0
* RabbitMQ server and client application log files
  - no entry found in log file.
* A runnable code sample, terminal transcript or detailed set of
  instructions that can be used to reproduce the issue
* RabbitMQ plugin information via `rabbitmq-plugins list`
  - rabbitmq_management
  - rabbitmq_management_agent
  - rabbitmq_web_dispatch
* Client library version (for all libraries used)
  - powershell
* Operating system, version, and patch level
  - Windows Server 2019 Standard

When a binding is created through Management API with routing_key: null, it is not possible to delete such binding through DELETE api call: /api/bindings/vhost/e/exchange/q/queue/props.

API returns 404 Not Found
```powershell
Invoke-RestMethod -Method Post -Uri https://rabbit01.hrebitom.poc:15671/api/definitions -Headers $headers -Body (  @{"exchanges" = @(@{"name" = "test"; "type" = "headers"; "vhost" = "/"; "durable" = $true; "auto_delete" = $false; "internal"= "false"; "arguments" = @{}}); "queues" = @(@{"name" = "test"; "vhost" = "/"; "durable" = $true; "auto_delete" = $false; "arguments" = @{}}); "bindings" = @( @{"source" = "test"; "destination" = "test"; "destination_type"= "queue"; "vhost" = "/"; "arguments" = @{} })} | ConvertTo-Json -Depth 10 )

Invoke-RestMethod -Method Get -Uri https://rabbit01.hrebitom.poc:15671/api/bindings/%2F/e/test/q/test -Headers $headers
# => 
# source           : test
# vhost            : /
# destination      : test
# destination_type : queue
# routing_key      : undefined
# arguments        :
# properties_key   : undefined

Invoke-RestMethod -Method Delete -Uri https://rabbit01.hrebitom.poc:15671/api/bindings/%2F/e/test/q/test/undefined -Headers $headers
# Invoke-RestMethod : {"error":"Object Not Found","reason":"Not Found"}
```
It's also not possible to delete the binding through Web UI.

Tom

michaelklishin:
I have a patch but I'm not convinced we should adopt it. There is no such thing as a "null routing key" in AMQP 0-9-1 clients or management UI form. There is such as thing as a blank routing key.

[...]

PR #725, opened 2019-07-27
 Parse routing keys with values of `null` and `undefined` to their respective atoms

michaelklishin:
See #723 for background.

Currently there are three main ways to declare a binding:

 * Use an AMQP 0-9-1 client
 * Use an HTTP API client
 * Definition import

For the sake of this change we will fold the latter two options into
one as they are effectively identical as far as data formats go.

An HTTP API client can do things that an AMQP 0-9-1 would not be able
to because of framing/parser/generator restrictions, and vice versa.

With HTTP API and definition import the following importance

 * When a binding is added, routing key can be omitted or set to `null` 
(or `undefined`)
 * When a binding is deleted, a blank routing key can only be specified
   using a convention since an empty URI path arguments are not 
supported
   by Cowboy REST and would make no sense to the developer.

Omitted or `null` `routing_key` values are parsed to their respective 
atoms. Those atoms are then stored in binding records. When a binding is 
retrieved or deleted (the case in #723), the conventionally encoded routing key 
that contained "null" for name was parsed to "null" (a string).
This  discrepancy led to false 404 responses from the API.

Note that this solution is merely a band aid over a fundamental 
difference in two APIs. Rejecting `null` or undefined values might be a better 
option but we cannot adopt it in 3.7.x. It remains to be discussed whether that
would improve user experience.

Another problematic scenario is routing keys with the value of "null"
(a string) used at binding time. This change would break deletion for
those. We'll assume that no developer would intentionally use such
routing key as the confusion and risk are more or less obvious.

### DIFF
--- a/src/rabbit_mgmt_wm_binding.erl
+++ b/src/rabbit_mgmt_wm_binding.erl
@@ -109,9 +109,16 @@ binding(ReqData) ->
 
 unpack(Src, Dst, Props) ->
     case rabbit_mgmt_format:tokenise(binary_to_list(Props)) of
-        ["~"]          -> {<<>>, []};
+        ["\~"]         -> {<<>>, []};
+        %% when routing_key is explicitly set to `null` in JSON payload,
+        %% the value would be stored as null the atom. See rabbitmq/rabbitmq-management#723 for details.
+        ["null"]       -> {null, []};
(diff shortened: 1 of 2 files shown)


michaelklishin:
Backported to `v3.7.x`.

Issue #85, opened 2015-11-13
Possible Cross site scripting in main.js


digennarot:
After I run security scan against rmq managment console i found this possible security issues:

Line 30:Unsafe client output calling window.location.replace() with tainted arg
Line 29:Assignment of "location" to user-controlled value
Line 29:Result of taint-preserving function call on user-controlled value
Line 29:"location.substr" is controlled by the user

Can we fix it ?

michaelklishin:
@digennarot that is an undocumented feature that allows logins from arguments in the URI hash, as opposed to filling out the form. Until we get OAuth 2 support, which will take a while, there is simply no alternative to this and certain products pretty much require this kind of login.

Take a look at the code, it doesn't do anything sophisticated. There is a worse aspect of it: that credentials have to be passed in the URI if you need to use this feature. But it is not required, not advertised anywhere, and OAuth 2 is the only realistic solution we know of.

So yes, we can fix it. By removing the need for it entirely.

smclewin:
For anybody finding this thread years later, a PR to address this can be found in #748

PR #722, opened 2019-07-23
Management UI can obtain an OAuth 2 token from UAA/CF SSO service

dcorbacho:
To use UAA authorization the rabbitmq-auth-backend-oauth2 must be enabled.
The oauth client must be configured with the scopes: `["openid", "rabbitmq.*]`

Follow the instructions to run UAA on that plugin:
https://github.com/rabbitmq/rabbitmq-auth-backend-oauth2#running-uaa

A configuration script is provided in this plugin on `scripts/seed.sh`,
to add an Oauth user for testing: `rabbit_user` with password: `rabbit_password`.

Management config:
```
[
 {enable_uaa, true},
 {uaa_client_id, "rabbit_user_client"},
 {uaa_location, "http://localhost:8080/uaa"}
]
```

[#158496811]

## Types of Changes
- [ ] Bugfix (non-breaking change which fixes issue #NNNN)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation (correction or otherwise)
- [ ] Cosmetics (whitespace, appearance)

## Checklist
- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [x] All tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in related repositories

PR #748, opened 2019-11-08
Remove direct login route

gerhard:
## Proposed Changes

This was an undocumented feature, but a potential XSS vector, as captured in #85. 

Since OAuth 2 support was introduced in #722, this functionality is less useful - it's time to remove it.

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #85)

## Checklist

## Further Comments

Remember to back-port to 3.8.x & 3.7.x

### DIFF
--- a/priv/www/js/dispatcher.js
+++ b/priv/www/js/dispatcher.js
@@ -14,7 +14,6 @@ dispatcher_add(function(sammy) {
             }
             render(reqs, 'overview', '#/');
         });
-    sammy.get('#/login/:username/:password', login_route);
 
     path('#/cluster-name', {'cluster_name': '/cluster-name'}, 'cluster-name');
     sammy.put('#/cluster-name', function() {
--- a/priv/www/js/main.js
+++ b/priv/www/js/main.js
@@ -44,31 +44,6 @@ function set_auth_pref(userinfo) {
     store_cookie_value_with_expiration('auth', encodeURIComponent(b64), date);
 }
 
-function login_route () {
-    var userpass = '' + this.params['username'] + ':' + this.params['password'],
-        location = window.location.href,
-        hash = window.location.hash;
-    set_auth_pref(decodeURIComponent(userpass));
-    location = location.substr(0, location.length - hash.length);
-    window.location.replace(location);
-    // because we change url, we don't need to hit check_login as
-    // we'll end up doing that at the bottom of start_app_login after
-    // we've changed url.
-}
-
-function login_route_with_path() {
-  var params = ('' + this.params['splat']).split('/');
-  var user = params.shift();
-  var pass = params.shift();
-  var userpass = '' + user + ':' + pass,
-        location = window.location.href,
-        hash = window.location.hash;
-    set_auth_pref(decodeURIComponent(userpass));
-    location = location.substr(0, location.length - hash.length) + '#/' + params.join('/');
-    check_login();
-    window.location.replace(location);
-}
-
 function getParameterByName(name) {
     var match = RegExp('[#&]' + name + '=([^&]*)').exec(window.location.hash);
     return match && decodeURIComponent(match[1].replace(/\+/g, ' '));
@@ -87,8 +62,6 @@ function start_app_login() {
             set_auth_pref(username + ':' + password);
             check_login();
         });
-        this.get('#/login/:username/:password', login_route);
-        this.get(/\#\/login\/(.*)/, login_route_with_path);
     });
     if (enable_uaa) {
         var token = getAccessToken();


michaelklishin:
@gerhard how would the PCF tile be adapted if we backport this to 3.8 and possibly 3.7? Do we know of any other internal users? It's technically a breaking change :/

smclewin:
Thank you for picking this up so quickly from my request to the security email alias.  I appreciate that you are considering back-porting the change as well.

gerhard:
As this wasn't a documented feature, I didn't think that it could be considered a breaking change. Let's discuss privately re PCF. I am not aware of anyone other than us using this feature.

gerhard:
We shared our intent to remove this functionality with the teams that we know will be affected by this change. If we don't hear a strong-enough case against merging this PR by the end of this week, the plan is to merge & ship it in the next `v3.8.x` patch (either v3.8.2 or v3.8.3 by the looks of it).

We will hold back-porting to `v3.7.x` for a while longer, so that if there is someone who really doesn't want losing this functionality can stay on a supported version of RabbitMQ for a couple more months. As a reminder, [3.7.x reaches end of life on 31 March 2020](https://www.rabbitmq.com/versions.html), after being in service for over 2 years.

michaelklishin:
We will decide on whether this can be backported to `v3.8.x` and `v3.7.x` in the next few weeks. There is agreement that this feature can go now but dependent project can take a long time to migrate to an alternative such as OAuth 2/JWT tokens.

michaelklishin:
This will be backported after `3.8.2` ships. Per discussion with @gerhard.

smclewin:
Good evening.  I would like to confirm this is in release 3.8.2.  I do not see mention of this particular change in the release notes for 3.8.2 (https://github.com/rabbitmq/rabbitmq-server/releases/tag/v3.8.2).  Given it was merged into master I assume it is in the release, but I'm looking for additional confirmation before I communicate that the change was incorporated to the customer of our product who reported this.   I've not downloaded the release, I've just read the release notes and did not find mention of issue #748.  @gerhard

michaelklishin:
@smclewin 3.8 releases are not produced off of master. See the milestone of this PR and my comment above:

```
This will be backported after 3.8.2 ships.
```

michaelklishin:
Backported to `v3.8.x` now that `3.8.2` is out.

carlhoerberg:
Seems like only UAA is supported for the mgmt interface redirect login, we would rather not remove this feature until there's another way to let our customers login to their servers without reverting to copy/pasting.

carlhoerberg:
I'm a bit worried that the redirect protocol for UAA isn't documented, and it's the only supported way to allow logins after this PR, except copy/pasting username and password, which no one wants. I guess we'll have to reverse engineer it...

smclewin:
Thank you for the clarification @michaelklishin.  I had not understood your release process and had told myself the story that the backporting you were talking about was into the 3.7.x line after this change came out in 3.8.2.  I get now that after 3.8.2 was released the backport from master (3.9) into the 3.8.x line would happen.  Thank you - I had simply not gotten that right.

From a release standpoint, it is a fair expectation that 3.8.3 would be the first formal release that would contain this patch, and that you generally work on a monthly release cadence?

michaelklishin:
@smclewin you are correct, it will ship in `3.8.3` in a few weeks (more likely than not in early January). There will be at least one preview release before so you can verify it for yourself, and [alpha releases](https://bintray.com/rabbitmq/all-dev) produced almost daily.

michaelklishin:
@carlhoerberg UAA docs are often not beginner-oriented but otherwise uses a relatively standard OAuth 2 flow (although every OAuth 2 tool/vendor has enough minor differences to be confusing at first) and a library that we now include. If you have specific questions, please post them to the mailing list.

carlhoerberg:
Could we add a POST path (eg. /login) that accepts username and password as form parameters, which sets the auth cookie and then redirect to / ? Reverse engineering UAA was easier said than done, certainly not standard oauth 2 flow. Also seems like `window.location` still is controllable by user input: https://github.com/rabbitmq/rabbitmq-management/blob/cc873bc4f8f1f0437016f952da93ccff7515809c/priv/www/js/main.js#L186

michaelklishin:
How would having a POST endpoint help with "automagic" login scenarios? The credentials would have to come from somewhere. Are you going to build a separate login form?

carlhoerberg:
our side, we have a form, where we generate the token or creds and then the form posts to the mgmt interface

```html
<form id="login" action="https://red-panda.rmq.cloudamqp.com/login" method="post">
  <input type="hidden" name="username" value="guest"/>
  <input type="hidden" name="password" value="guest"/>
</form>  
<script>document.forms["login"].submit()</script>
```

michaelklishin:
OK, this might be a good middle ground. Let us discuss. @gerhard

gerhard:
The `POST /login` proposed behaviour sounds most sensible, I cannot think of any reason why we shouldn't/couldn't implement it.

Thanks for spotting the other instance of `window.location` which is susceptible to user input tampering, we should definitely look at that next.

I won't be able to pick this up for a few weeks, contributions are welcome.

smclewin:
> 
> 
> @smclewin you are correct, it will ship in `3.8.3` in a few weeks (more likely than not in early January). There will be at least one preview release before so you can verify it for yourself, and [alpha releases](https://bintray.com/rabbitmq/all-dev) produced almost daily.

@michaelklishin does Pivotal have an updated sense of when `3.8.3` is intended to be released now that early January is behind us?  I'm asking to help manage both our own expectations and that of the customer that originally brought the XSS vulnerability to our attention in late October.

michaelklishin:
Next month.

PR #750, opened 2019-11-12
Move definition import machinery to core

### LINKED ISSUE #749: Move definition import implementation into the core
With [Prometheus and Grafana support](https://www.rabbitmq.com/prometheus.html) in 3.8 some users would like to disable this plugin but it's not currently possible to export and import definitions without it.

The definition import modules are tied to this plugin for historical reasons. There is no reason why this feature cannot be available, say, via `rabbitmqctl` on a stock node. So let's make that possible.

The modules in this plugin would forward to the core ones, and the API endpoints will remain.
The goal is to keep everything backward compatible while making it possible to introduce a `rabbitmqctl import_definitions` or a similar command(s).

michaelklishin:
## Proposed Changes

This moves definition import machinery and related helpers to the core while retaining some exported functions for backward compatibility.

Depends on

 * https://github.com/rabbitmq/rabbitmq-server/pull/2160
 * https://github.com/rabbitmq/rabbitmq-common/pull/339
 * https://github.com/rabbitmq/rabbitmq-cli/pull/387

## Types of Changes

## Checklist

## Further Comments

Closes #749.


michaelklishin:
Backported to `v3.8.x`.

PR #755, opened 2019-11-20
pass authorisation context instead of channel source

velimir:
Part of rabbitmq/rabbitmq-server/pull/2169


michaelklishin:
Backported to `v3.8.x`.

Issue #304, opened 2016-11-30
Deleting a binding with `rabbitmqadmin` with empty properties key should be clearer or documented


s0rc3r3r01:
Deleting a binding with an empty properties_key should be correctly documented or described.

The behaviour seems no to be described anywhere, the way I managed to do it, is by usign _"~"_  as properties_key

`rabbitmqadmin delete binding source=XX destination=YY destination_type=queue properties_key="~" --username=YY --password=XX`

michaelklishin:
Are you sure that the tilde wasn't expanded to `$HOME` in the above example (that is, the key is that the value wasn't actually empty)?

s0rc3r3r01:
hi Micheal, yes sure, it looks pretty unlogical to have my home folder as properties_key...

michaelklishin:
OK, I'll assume this was a shell expansion artifact then.

fozboz:
It isn't.
```
$ rabbitmqadmin declare binding source=anexchange destination=aqueue destination_type=queue
binding declared
$ rabbitmqadmin list bindings source destination routing_key properties_key | grep anexchange
| anexchange | aqueue    |             | ~              |
```
Same happens when creating a binding with the Web UI. When a routing key is specified, the properties key gets set to the same value.
What is the properties key for and why is it needed to delete a binding?

michaelklishin:
It is needed because it's a field on both `{queue,exchange}.bind` and `{queue,exchange}.unbind` protocol methods. You can set it to an empty string. I agree that it should be the default but when it comes to the HTTP API and `rabbitmqadmin`, it's complicated because it's an arbitrary map of keys and values that has to be an endpoint path segment, so it has to be/will be encoded.

`rabbitmqadmin` is a zero dependency Python script. Feel free to take a look and suggest changes via pull requests. They will be considered.

IvanKaloyanov:
Hi, [this](https://github.com/rabbitmq/rabbitmq-management/pull/779) is now merged so as it is for the create binding, now you don't have to explicitly set value for `properties_key` when you want to delete a binding with empty `properties_key`

michaelklishin:
And #779 will ship in `3.8.3`.

PR #779, opened 2020-02-29
properties_key default value on binding deletable

IvanKaloyanov:
Align create binding and delete banding convention. Use an empty
'~' properties_key by default as multiple users and I personally
find it inconvenient (and at first really time consuming)
to use the $HOME symbol for unbinding the default/simplest binding.

## Proposed Changes

Hi team,

This week I needed to troubleshoot RMQ relate code on an AWS ec2 instance and used `rabbitmqadmin` tool to do so. I encounter that when you create a simple queue/exchange binding:
`$ ./rabbitmqadmin declare binding source=myExchange destination='myQueue'`
The binding is created with `routing_key=''` and `properties_key='~'` by default. Then when I try to delete the newly created binding there is something that I find inconvenient and that's the fact that I have to explicitly set the `properties_key` value even if it is the one that is set by default - '\~'. I think that the two operations (declare & delete binding) should follow the same convention and the delete action should use '~' as a default value for `properties_key` as well. On the other hand, I and from what I can tell a few other people [issue 304](https://github.com/rabbitmq/rabbitmq-management/issues/304) find the deletion of the default/empty `properties_key` "hard to guess" so to speak as the most convenient and the things you try at first (delete without `properties_key` or `properties_key=''`) does not work so I needed to dive into the code to see what's going on.

Hopefully, you find this a beneficial change as I think this will save a lot of people's time when they want to delete a binding without properties_key.

Thanks,
Ivan
## Types of Changes

## Checklist

- [x] I have added necessary documentation (if appropriate)

### DIFF
--- a/bin/rabbitmqadmin
+++ b/bin/rabbitmqadmin
@@ -165,8 +165,8 @@ DECLARABLE = {
 DELETABLE = {
     'exchange':   {'mandatory': ['name']},
     'queue':      {'mandatory': ['name']},
-    'binding':    {'mandatory': ['source', 'destination_type', 'destination',
-                                 'properties_key']},
+    'binding':    {'mandatory': ['source', 'destination_type', 'destination'], 
+                    'optional': {'properties_key': '~'}},
     'vhost':      {'mandatory': ['name']},
     'user':       {'mandatory': ['name']},
     'permission': {'mandatory': ['vhost', 'user']},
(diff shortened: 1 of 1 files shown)


michaelklishin:
Since the key is currently required this apparently won't be a breaking change, which is always welcome.

michaelklishin:
I agree with your thinking. Thank you.

michaelklishin:
Backported to `v3.8.x`.

michaelklishin:
Backported to `v3.7.x`.

PR #805, opened 2020-04-18
rabbitmqadmin: make --cli-switches take precedence over config file values

### LINKED ISSUE #804: .rabbitmqadmin.conf settings take precedence over CLI switches
I've found that the `rabbitmqadmin` command was not respecting the `--host` switch on the command line and instead was using the `[default]` `hostname` field's value, leading to confusion over socket timeouts.

Usually explicit command line switches should take priority over implicit values like the default values of a configuration file, or environment variables.

Example configuration file:

``` ini
[default]
hostname = unreachable-host
port = 15672
```

and an example invocation:

``` shell
./bin/rabbitmqadmin -c /path/to/configs/issues/rabbitmqadmin.management-804.conf --host localhost list users
```

michaelklishin:
## Proposed Changes

rabbitmqadmin: make --cli-switches take precedence over config file values

With most CLI tools, command line arguments
take precedence over values in the configuration file.

This was not the case in rabbitmqadmin, and very likely
unintentionally so, at least I could not find
any evidence of the contrary.

There was a test case that implicitly depended
on this behavior. Again, no indication of this
being an intentional design choice.

While this can be a breaking change, most
users either use CLI flags or the config file;
this is why this behavior has gone unnoticed
for at least 8 years. We therefore treat
this change as low risk and worth
shipping e.g. in a patch release.
rabbitmqadmin is installed manually and
therefore won't be replaced during a node
upgrade anyway. Operators would
have to opt-in.

## Types of Changes

- [ x Bugfix (non-breaking change which fixes issue #804)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x] I have added necessary documentation (if appropriate)

## Further Comments

Closes #804.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ variables-ct*
 *.coverdata
 
 test/config_schema_SUITE_data/schema/
+.vscode/*
--- a/bin/rabbitmqadmin
+++ b/bin/rabbitmqadmin
@@ -20,6 +20,7 @@ from __future__ import print_function
 from optparse import OptionParser, TitledHelpFormatter
 
 import base64
+import copy
 import json
 import os
 import socket
@@ -331,6 +332,8 @@ def fmt_usage_stanza(root, verb):
 
 default_options = {"hostname": "localhost",
                    "port": "15672",
+                   # default config file section name
+                   "node": "default",
                    "path_prefix": "",
                    "declare_vhost": "/",
                    "username": "guest",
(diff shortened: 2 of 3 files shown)


michaelklishin:
@HariSekhon can you please give `rabbitmqadmin` from this branch a try and confirm that it behaves the way that meets your expectations?

michaelklishin:
Backported to `v3.8.x`.

Issue #774, opened 2020-02-24
[improvement] Add permissions required for API routes and/or migrate to Swagger


artandor:
As explained in the title, the current documentation lack of informations about which permissions are required to call certain API routes. It could be a huge improvement for users to just need a quick look at a route to know what are the requirements.

It could be done in the current table present at localhost:15672/api, but it seems to me that this UI could be improved a lot by providing a Swagger UI. That would allow what i previously said, and also allow a quick display of possible parameters AND a quick way to test a route.

artandor:
Is the file to update [this one](https://github.com/rabbitmq/rabbitmq-management/blob/master/priv/www/api/index.html) ?  @michaelklishin 
Is there a specific place that those infos would be all referenced ?

Considering the change of title i guess including swagger UI is off topic for now ?

michaelklishin:
We'd consider a contribution of a Swagger reference. On this team, one issue equals one actionable item, not two or more.

The page is under [priv/www](https://github.com/rabbitmq/rabbitmq-management/tree/master/priv/www). Every [endpoint module](https://github.com/rabbitmq/rabbitmq-management/tree/master/src) has a function that determines if the caller has sufficient HTTP API permissions, here's [one example](https://github.com/rabbitmq/rabbitmq-management/blob/master/src/rabbit_mgmt_wm_bindings.erl#L111). Protocol method permissions are listed in a table [in the docs](https://www.rabbitmq.com/access-control.html).

michaelklishin:
Some HTTP API endpoints are not tied to any respective AMQP 0-9-1 methods, so their permissions come down to whether the user is tagged appropriately. Others are tied to protocol methods, in which case their permission docs should be taken from the [main permission table](https://www.rabbitmq.com/access-control.html).

artandor:
Thanks for the informations. Considering that the permissions informations are needed before trying to implement a swagger ui, i think i'll focus on updating the api documentation if I find the time to contribute.

artandor:
Hey, i started doing the changes to update the doc. I'd like to know if my "template" is right, since if it's not i'd have to start all over again.

You can find the updated code here : https://github.com/artandor/rabbitmq-management/blob/document_permissions_api/priv/www/api/index.html

Globaly i was planning on following this pattern :
<required tag> & [Method] <required tag> / <required vhost permissions>

I'll wait for your answer before I spend more time.

Thanks

artandor:
[...]

PR #807, opened 2020-04-22
Document permissions api

artandor:
## Proposed Changes

This PR try to answer the issue exposed in #774 
I documented the permissions required for every API route in order to have a clear overview of the required permissions when checking the documentation.

## Types of Changes

- [x] Documentation (correction or otherwise)

## Checklist

### DIFF
--- a/priv/www/api/index.html
+++ b/priv/www/api/index.html
@@ -175,6 +175,7 @@ vary: accept, accept-encoding, origin</pre>
         <th>DELETE</th>
         <th>POST</th>
         <th>Path</th>
+        <th>Required permissions (API / vhost)</th>
         <th>Description</th>
       </tr>
       <tr>
@@ -183,6 +184,7 @@ vary: accept, accept-encoding, origin</pre>
         <td></td>
         <td></td>
         <td class="path">/api/overview</td>
+        <td>management</td>
         <td>Various random bits of information that describe the whole
         system.</td>
       </tr>
@@ -192,6 +194,7 @@ vary: accept, accept-encoding, origin</pre>
         <td></td>
         <td></td>
         <td class="path">/api/cluster-name</td>
+        <td>management & [PUT] administrator</td>
         <td>Name identifying this RabbitMQ cluster.</td>
       </tr>
       <tr>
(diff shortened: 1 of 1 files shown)


priv/www/api/index.html
```diff
@@ -403,6 +443,10 @@ <h2>Reference</h2>
         <td>X</td>
         <td></td>
         <td class="path">/api/exchanges/<i>vhost</i>/<i>name</i></td>
+        <td>management</td>
+        <td>User added to vhost<br/>
```
michaelklishin, line 447: A user cannot be "added to a vhost". A user can have no or some permissions in a vhost.

priv/www/api/index.html
```diff
@@ -403,6 +443,10 @@ <h2>Reference</h2>
...
+          [PUT & DELETE] Configuration
```
michaelklishin, line 448: "Configuration" is not a role or permission in RabbitMQ. `configure` is a permission (for a single virtual host).

priv/www/api/index.html
```diff
@@ -437,6 +485,8 @@ <h2>Reference</h2>
         <td></td>
         <td>X</td>
         <td class="path">/api/exchanges/<i>vhost</i>/<i>name</i>/publish</td>
+        <td>management</td>
+        <td>User added to vhost & write permission</td>
```
michaelklishin, line 489: This should be clear as to what the `write` permission is on: the target exchange.

michaelklishin (changes requested):
Thank you for your time. This needs some changes before it would be considered. Specifically:

 * Nearly all whitespace changes are not relevant and it would be great to undo them
 * "A user added to a vhost" is neither specific nor a correct term. A user can have some permissions in a vhost. In which case we should specify what they are.
 * "Administrator or policymaker" should be just "policymaker" since "administrator" includes "policymaker" just like "policymaker" is "management" plus the ability to operate on policies
 * Please changes `[PUT & DELETE]` to `PUT`, `DELETE`
 * It may be worth having "Tags: …" and "object permissions: …" instead of "administrator or policymaker & added to vhost" (free form in a single line, plus not specific enough)

dcorbacho:
@artandor The management plugin has been moved to the [server](https://github.com/rabbitmq/rabbitmq-server) repository. Please re-submit your PR against this repository once you have applied the requested changes.

Many thanks for your contribution!

PR #809, opened 2020-04-30
GitHub Actions

pjk25:
## Proposed Changes

Run the test suite via Actions on each push

## Types of Changes

- [x] Testing

## Further Comments

Workflow was generated using `make github-actions`, and remove the CHECK SCRIPTS step that would invoke `make bats`.

### DIFF
--- a/.github/workflows/test-erlang-otp-21.3.yaml
+++ b/.github/workflows/test-erlang-otp-21.3.yaml
@@ -1,0 +1,547 @@
+# https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow
+name: "Test - Erlang 21.3"
+on: push
+jobs:
+  checks:
+    name: checks
+    runs-on: ubuntu-18.04
+    steps:
+      - name: CHECKOUT REPOSITORY
+        uses: actions/checkout@v2
+      # https://github.com/marketplace/actions/setup-elixir
+      - name: CONFIGURE OTP & ELIXIR
+        uses: actions/setup-elixir@v1
+        with:
+          otp-version: 21.3
+          # https://github.com/elixir-lang/elixir/releases
+          elixir-version: 1.8.0
+      - name: CHECK RABBITMQ COMPONENTS
+        # https://github.community/t5/GitHub-Actions/How-can-I-set-an-expression-as-an-environment-variable-at/m-p/41804/highlight/true#M4751
+        id: ref
+        run: |
+          branch_or_tag_name=${GITHUB_REF#refs/*/}
+          echo "::set-output name=branch_or_tag_name::$branch_or_tag_name"
+          make check-rabbitmq-components.mk base_rmq_ref=master current_rmq_ref=$branch_or_tag_name
+      # https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows
+      - name: CACHE DEPS
+        uses: actions/cache@v1
+        with:
+          path: deps
+          key: otp-21.3_git-${{ github.sha }}_deps
+      - name: RESOLVE & COMPILE DEPS
+        run: |
+          make deps test-deps base_rmq_ref=master current_rmq_ref=${{ steps.ref.outputs.branch_or_tag_name }}
+          echo "Capture versions of the RabbitMQ components used in this workflow..."
+          make rabbitmq_management-rabbitmq-deps.mk
+          mv rabbitmq_management-rabbitmq-deps.mk deps/
+          echo "Remove directories not used in the subsequent jobs..."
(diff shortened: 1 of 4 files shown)


.github/workflows/test-erlang-otp-21.3.yaml
```diff
@@ -0,0 +1,547 @@
...
+          path: deps
+          key: otp-21.3_git-${{ github.sha }}_deps
+      - name: FORMAT GIT REF
+        # https://github.community/t5/GitHub-Actions/How-can-I-set-an-expression-as-an-environment-variable-at/m-p/41804/highlight/true#M4751
+        id: ref
+        run: |
+          branch_or_tag_name=${GITHUB_REF#refs/*/}
+          echo "::set-output name=branch_or_tag_name::$branch_or_tag_name"
+      - name: UPLOAD TO S3
+        if: github.ref == 'refs/heads/master'
```
gerhard, line 538: Let's comment this (for both workflows) and double-check that it works before we uncomment and merge.

gerhard:
Apart from the missing secrets required for the S3 upload (I've just added them now), it all looks good to me. I've left an inline comment re this.

Does this look good to you @dumbbell ?

gerhard:
Oh, and let's add some badges via shields.io, e.g. https://github.com/rabbitmq/rabbitmq-prometheus/commit/d0378011a986bdd4077000b3a8d141b30ca6c41b

pjk25:
@gerhard There is some flakiness to the rabbit_mgmt_test_db suite. Before we add the badges I'd like to see if I can fix that.

pjk25:
@gerhard I've encountered the test flake locally when running the whole suite, so there is something worth fixing, although maybe not as part of this PR.

As for removing Travis, I think we should go ahead since we have equivalent coverage with Actions. It's not as if running both improves our confidence - if anything it could add confusion, and it's slower.

And let's use the GitHub badges rather than have a workaround. It's easy to change the name, but then we may hit that issue again later.

pjk25:
They failure on Travis is reproducible locally when you run the whole suite. Afterwards, if I `ps aux | grep erl`, it looks like a test isn't cleaning up correctly:
```
paralle+ 16435  1.2  0.5 3244044 89984 ?       Sl   09:37   0:08 ct_run -W w -K true -A 64 -MBas ageffcbf -MHas ageffcbf -MBlmbcs 512 -MHlmbcs 512 -MMmcs 30 -P 1048576 -t 5000000 -stbt db -zdbbl 128000 -S 2 -sbwt very_short -A 24 -B i -- -root /usr/lib/erlang -progname erl -- -home /home/parallels/workspace/github.com/rabbitmq/rabbitmq-public-umbrella/deps/rabbitmq_management/logs/ct_run.ct_rabbitmq_management@parallels-Parallels-Virtual-Platform.2020-05-01_09.24.21/deps.rabbitmq_management.rabbit_mgmt_rabbitmqadmin_SUITE.logs/run.2020-05-01_09.37.08/log_private/ -- -pa  -noshell -noinput -s rabbit boot -boot start_sasl -ra wal_sync_method sync -lager crash_log false -lager handlers [] --
paralle+ 16524  0.0  0.0   4532  1512 ?        Ss   09:37   0:00 erl_child_setup 1024
paralle+ 19593  0.0  0.0  21536  1004 pts/0    S+   09:49   0:00 grep --color=auto erl
```

Which I think tripped me up on a different day, so I will see if I can fix it.

gerhard:
> @gerhard I've encountered the test flake locally when running the whole suite, so there is something worth fixing, although maybe not as part of this PR.

👍🏻
 
> And let's use the GitHub badges rather than have a workaround. It's easy to change the name, but then we may hit that issue again later.

The lack of proper alignment is the reason why GitHub Actions badges are a turn-off for me:

image

This is what we (+@dumbbell) did for rabbitmq-server https://github.com/rabbitmq/rabbitmq-server/commit/cb99072d7e9f6ffa5bbfe0be78a1cf5c587942a3, and this is the end-result:

image

Notice that we have changed the workflow name to e.g. `Test - Erlang 21.3`

pjk25:
Alignment was actually fine in the IntelliJ preview, which is interesting. I'll switch it back over and rename the workflow.

pjk25:
@dumbbell fixed the flakiness with https://github.com/rabbitmq/rabbitmq-management/commit/9100530f9884239b2edf93277a16934c3174f573, and I have cleaned up the branch history, so I think this is now good to go @gerhard

michaelklishin:
@gerhard should we merge this?

gerhard:
@michaelklishin a merge is definitely long overdue. As soon as the checks that are currently running pass, I will merge.

PR #811, opened 2020-05-06
Initial Honeycomb Integration

HoloRin:
## Proposed Changes

Collect test statistics with Honeycomb.io. Captures 

- Github Workflow
- Github Run Id
- Github Repository
- Github Sha
- Github Ref
- Test Suite Name
- Test Case Name
- Test Case Duration
- Test Case Result

only at this time, but this should be enough to perform some queries in Honeycomb to identify flaky and/or slow tests.

## Types of Changes

- [x] Continuous Integration

### DIFF
--- a/.github/workflows/test-jobs/30-CT_SUITE.yaml
+++ b/.github/workflows/test-jobs/30-CT_SUITE.yaml
@@ -29,10 +29,10 @@ ct-$(CT_SUITE):
           FULL= \
           FAIL_FAST=1 \
           SKIP_AS_ERROR=1 \
-          CT_OPTS="-ct_hooks honeycomb_cth"
+          CT_OPTS="-ct_hooks honeycomb_cth '[{directory,\"$PWD/honeycomb\"}]'"
     - name: HONEYCOMB
       run: |
-        for f in /tmp/honeycomb/*; do
+        for f in honeycomb/*; do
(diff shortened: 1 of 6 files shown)


pjk25:
Only the last commit matters, as this builds off of #809 which is not merged yet.

.github/workflows/test-jobs/30-CT_SUITE.yaml
```diff
@@ -0,0 +1,77 @@
...
+            base_rmq_ref=$(base_rmq_ref) \
+            current_rmq_ref=$branch_or_tag_name \
+            FULL= \
+            FAIL_FAST=1 \
+            SKIP_AS_ERROR=1 \
+            CT_OPTS="-ct_hooks honeycomb_cth '[{directory,\"$PWD/honeycomb\"}]'"
+      - name: HONEYCOMB
+        run: |
+          for f in honeycomb/*; do
+            curl -H 'X-Honeycomb-Team: ${{ secrets.HONEYCOMB_TEAM }}' \
```
pjk25, line 39: This should receive the same return code validation as was added to rabbitmq-cli, and probably also move to the tail of the steps, so that if it fails, it does not affect calls to `success()` in the `if:` of the mixed versions tests.

Issue #818, opened 2020-06-09
Respect management.load_definitions for backwards compatibility


michaelklishin:
Ever since definition load on boot has moved into the core, `management.load_definitions` is unintentionally ignored. Instead of trying to work around this in the core we should do it in this plugin, e.g. with an additional boot step.

PR #819, opened 2020-06-09
Load definitions from rabbitmq.load_definitions via a boot step

michaelklishin:
Previously we tried to rely on `rabbitmq_management` env
being loaded in the core. That is suboptimal and has unintentional side effects.

References rabbitmq/rabbitmq-management#2368, rabbitmq/rabbitmq-management#2373.
Closes rabbitmq/rabbitmq-management#818.


michaelklishin:
Backported to `v3.8.x`.

Issue #821, opened 2020-06-10
Make it possible to configure rabbitmq_management.management_db_cache_multiplier via rabbitmq.conf


michaelklishin:
There is no reason why this setting should not be configurable via `rabbitmq.conf`

PR #822, opened 2020-06-10
Make it possible to configure db cache multiplier via rabbitmq.conf

michaelklishin:
Closes rabbitmq/rabbitmq-management#821.


michaelklishin:
Backported to `v3.8.x`.

PR #837, opened 2020-08-24
fix zero time formatted as empty string

luos:
## Proposed Changes

Previously the check was using non-strict equality, which meant that empty
string was displayed instead of 0. This can mislead users to think that
no value is set while actually, for example, the federation message expiry is
set to 0.

I checked other repos as well, and it always seems to be used the same way by the connection page and the shovel plugin. No other usages found. 

Here is a picture of the changes on the federation page: 

image

## Types of Changes

## Checklist

### DIFF
--- a/priv/www/js/formatters.js
+++ b/priv/www/js/formatters.js
@@ -72,7 +72,7 @@ function fmt_timestamp_mini(ts) {
 }
 
 function fmt_time(t, suffix) {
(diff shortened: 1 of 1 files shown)


michaelklishin (approved):

michaelklishin:
Thank you!

michaelklishin:
Backported to `v3.8.x`.

PR #849, opened 2020-09-24
GitHub Actions: test on OTP 22.3 and 23

michaelklishin:
This now depends on #850 since the plugin currently does not pass `gmake dialyze`.


dumbbell:
I'm not the author of this GitHub Actions workflow, so I may not be the best to review it.

michaelklishin:
This has revealed a lot of Dialyzer failures, a few seem legit (but potentially handled by forgiving Cowboy functions).

